PR #4, opened 2015-06-05
Add semver config files

tbetbetbe:
- Adds configuration for the semantic version to use in packages being generated from the services.

### DIFF
--- a/google/example/library/v1/semver.yml
+++ b/google/example/library/v1/semver.yml
@@ -1,0 +1,15 @@
+# This configures the semantic version to be used in packaging this api.
+#
+# E.g,
+#
+# default: 0.9.0  # means that package manager should use 1.0.0 unless there
+#                 # is a more specific version specifed
+#
+# For most purposes this should be sufficient, However, it may be necessary to
+# have re-release the packages, so each language to specify an override.
+#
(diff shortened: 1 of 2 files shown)


wora:
I am not sure why this is needed. The major api version is declared in the proto package name. The patch indicates server bug fixes, which don't require client changes. The question is where do we get the minor version from. I think we need to find a way for API owners to declare such version. We, as repo maintainers, should not assign the versions.

tbetbetbe:
- the major version of any semver-released package needs to be able to vary independently of the api version.
  - e.g today google-apis-pubsub-python is released at 1.0.0 with along grpc-python.0.9.0a3
  - grpc-python updates with an api change grpc-python-1.1 which contains a breaking api change
  - google-apis-pubsub-python needs to be re-released
     - it cannot use semver 1.x.x even though the proto api has not changed, it must now be 2.0.0

wora:
For each API, we should generate api package with name like grpc-google-pubsub-v1, and the dependencies contribute to the package version, such as 0.9.0. But the package version should be consistent across all APIs. It should be configured centrally, not per API. Right?

bjwatson:
@tbetbetbe Can this be closed?

guptasu:
Seems like this is now out of date. As per request from @wora , let's close this PR for the sake of cleanup and we can create another one when we need to add symver information.

PR #5, opened 2015-06-05
Add package_deps.yml

tbetbetbe:
Configures the semantic versions of the dependencies to use when generating 
googleapis packages from the api protos

### DIFF
--- a/package_deps.yml
+++ b/package_deps.yml
@@ -1,0 +1,27 @@
+# This configures the version of the grpc and other packages 
+# (e.g, the google auth library) to be used as dependencies 
+# of the googleapis packages generated from the protos in
+# this repo.
+#
+# Rationale:
+# - at a given point, in any language, googleapis packages generated should depend
+#   on consistent versions of their dependencies
+# - the tool that creates the packages should pick up the version from a configuration file
+# - it's nice to have the information all in one place to make it convenient to update
+#   e.g, when new apis are added or when there is a gRPC release.
+
+csharp:
+  grpc: 0.5.1
+
+python: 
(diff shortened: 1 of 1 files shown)


wora:
I think this file should be auto generated by the release process and be part of the output in another repo. The source repo should not need to have these information. The general concept is people can use this repo to generate any client/server/docset as they need, we don't need to dictate any dependency.

tbetbetbe:
> and be part of the output in another repo

Which one?  At the moment, this is the only repo needed when generating gRPC-based packages.
We could put the generated file in another repo, but then the tool that builds the packages needs to get its info from two sources.

IMHO, its simpler for tools (and humans) if it lives here.
- tools doing any processing get all their info from one source and there is minimal chance of inconsistencies occurring
- Non-grpc tools that don't need the dependency info can just ignore its contents.

wora:
This looks good. Do you want to update the version numbers before I merge it?

tbetbetbe:
Sorry, about the lag in the response - I will update the version numbers in a commit later today.

garrettjonesgoogle:
It seems that this has gotten somewhat out of date in comparison to https://github.com/googleapis/packman/blob/master/config/dependencies.yml .

At a higher level, the toolkit team as agreed that it would be a good idea to move the versions from packman into googleapis. This will require a coordinated change across multiple repositories (packman/artman/googleapis). Maybe the present PR should be closed and a new effort should be taken up for moving the version config.

guptasu:
As per request from @wora , seems like this is now out of date. Let's close this PR for the sake of cleanup and we can create another one when we need to add symver information.

PR #6, opened 2015-07-13
Syncs the google well-known protos with protobuf-3.0.0-alpha-3.1

tbetbetbe:
- alpha users of gRPC are using an alpha version of protoc
- its breaks with the `option javanano_use_deprecated_package = true;` option as this is not recognized
- it includes official support for C# and objective-c

### DIFF
--- a/google/protobuf/any.proto
+++ b/google/protobuf/any.proto
@@ -1,17 +1,32 @@
-// Copyright (c) 2015, Google Inc.
+// Protocol Buffers - Google's data interchange format
+// Copyright 2008 Google Inc.  All rights reserved.
+// https://developers.google.com/protocol-buffers/
 //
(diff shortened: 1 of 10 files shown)


tbetbetbe:
@wora, PTAL

Rationale:
At the moment the [in-progress packaging tool](https://github.com/grpc/grpc-tools/tree/master/lib) breaks when it tries to build packages directly from the GitHub using the latest alpha release of protoc because of the unrecognized option.

The changes here sync theses protos to the google protobuf v3.0.0.alpha-3.1 tag which fixes that issue

We should probably decide on a strategy for keeping these copies of the well-known protos up-to-date until proto3 is officially released.

wora:
Can you also copy over descriptor.proto with another PR? That way, everything is self-contained.

PR #16, opened 2016-05-03
Add cloudtrace gapic config

geigerj:


geigerj:
FYI: @jmuk

gapic/api/artman_cloudtrace.yaml
```diff
@@ -0,0 +1,16 @@
...
+  service_yaml:
+    - ${THISDIR}/../../google/devtools/cloudtrace/v1/trace.yaml
+  gapic_api_yaml:
+    - ${THISDIR}/../../google/devtools/cloudtrace/v1/trace_gapic.yaml
+  auto_merge: true
+  auto_resolve: true
+  ignore_base: false
+  output_dir: ${REPOROOT}/artman/output
+cloudtrace_python:
+  final_repo_dir: ${REPOROOT}/artman/output/gcloud-python-cloudtrace
```
jmuk: Can you add ruby config that is
```
cloudtrace_ruby:
  final_repo_dir: ${REPOROOT}/artman/output/gcloud-ruby-cloudtrace
```

?

google/devtools/cloudtrace/v1/trace_gapic.yaml
```diff
@@ -0,0 +1,66 @@
+type: io.gapi.vgen.ConfigProto
+language_settings:
+  python:
+    package_name: google.devtools.cloudtrace.v1
```
jmuk: Similarly, could you add
```
ruby:
  package_name: Google::Devtools::Cloudtrace::V1
```
?

geigerj:
I've added the Ruby config, but the CI is now failing. Do you know why? Appears to be Ruby-related

jmuk:
I don't know why CI fails (actually it failed before I put my comments above). Honestly I don't know why travis wants to run something even though .travis.yml does not exist.

I think you can simply ignore it. I turned off the travis config in travis-ci.org for this project, I hope that change works well.

google/devtools/cloudtrace/v1/trace_gapic.yaml
```diff
@@ -0,0 +1,68 @@
...
+  ruby:
+    package_name: Google::Devtools::Cloudtrace::V1
+interfaces:
+- name: google.devtools.cloudtrace.v1.TraceService
+  collections:
+  - name_pattern: projects/{project}/traces/{trace_id}
```
tbetbetbe: Having chatted to the cloud trace team, you can remove this as it's misleading.
In the trace api, the ids are just the raw ids, and not resource names

garrettjonesgoogle: wat

geigerj: Done.

(I think this is correct; if you look at the HTTP bindings in trace.proto, there is no parameter bound to the whole path, just the individual segments.)

tbetbetbe:
LGTM

gapic/api/artman_cloudtrace.yaml
```diff
@@ -0,0 +1,18 @@
+cloudtrace_common:
```
jmuk: Since I've committed #22, these keys would be 'common', 'python', and 'ruby' -- no 'cloudtrace_' prefix is necessary.

geigerj: Fixed

PR #23, opened 2016-05-10
Update GAPIC configs

geigerj:
Corresponds to the toolkit refactoring, googleapis/toolkit#71


garrettjonesgoogle:
LGTM

bjwatson:
LGTM2

It occurs to me that we might move language_provider up a level again, because it no longer varies between templates. Just a thought for a future PR.

geigerj:
Agreed -- let's file that as an issue in the toolkit repo.

bjwatson:
@geigerj Okay, let's see how it plays out with language_provider being at the template level.

PR #27, opened 2016-05-11
Add the config for remote pipeline

jmuk:
This is meaningful only with https://github.com/googleapis/artman/pull/22

### DIFF
--- a/gapic/api/artman_common.yaml
+++ b/gapic/api/artman_common.yaml
@@ -1,0 +1,25 @@
+common:
+  import_proto_path:
+    - ${THISDIR}/../..
+  src_proto_path:
+    - ${THISDIR}/../../google/${API_VERSION}
+  service_yaml:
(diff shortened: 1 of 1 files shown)


gapic/api/artman_common.yaml
```diff
@@ -0,0 +1,25 @@
+common:
```
garrettjonesgoogle: If we can depend on this file, why not get rid of the individual config files with the hardcoded values?

jmuk: execute_pipeline.py is a generic utility and does not have good commandline to allow expanding variables like 'API_VERSION' or 'API_NAME'. My new commandline tool, api_codegen.py, has narrower focus and therefore can use this simplified format.

Until execute_pipeline.py is redesigned to allow this config, I think it's better to keep the individual API config files.

bjwatson:
@jmuk Is this PR obsolete?

jmuk:
Right.

PR #34, opened 2016-05-21
Add required_fields to logging_gapic config

jmuk:
Noticed we do not have required_fields in logging API...

### DIFF
--- a/google/logging/v2/logging_gapic.yaml
+++ b/google/logging/v2/logging_gapic.yaml
@@ -41,6 +41,8 @@ interfaces:
       groups:
       - parameters:
         - project_name
+    required_fields:
+      - project_name
     page_streaming:
       request:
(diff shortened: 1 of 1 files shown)


google/logging/v2/logging_gapic.yaml
```diff
@@ -82,6 +89,9 @@ interfaces:
       - parameters:
         - sink_name
         - sink
+    required_fields:
+      - sink_name
+      - sink
```
garrettjonesgoogle: (I can't seem to leave a comment on unmodified code) I think DeleteSink's parameter sink_name is required.

jmuk: done, thanks for catching!

garrettjonesgoogle:
LGTM

PR #40, opened 2016-06-06
Go package

saicheems:


jmuk:
If you're thinking to add them in another CL, then that's fine, this CL looks good to me by itself.

saicheems:
Right, I was going to go one package at a time.
I'll hold off for a bit anyway, seems like there might be some indecision on the import paths again...

PR #47, opened 2016-06-16
Python release fixes

bjwatson:
This includes:

* renaming the Python package names for package delivery, which is what we're doing right now
* adding service-level comments, since Sphinx depends on class-level docstrings in the generated code (which come from service-level proto comments)
* fixing a typo in the proto comments (also fixed upstream in a CL)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@ OWNERS
 README.google
 google/internal
 google/protobuf
+.project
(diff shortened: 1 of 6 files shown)


tbetbetbe:
LGTM.

Do we have a process in place for making sure the proto comments are reflected internally ?

anthmgoogle:
LGTM.

bjwatson:
@tbetbetbe I can ping the logging team with my proposed comments, which they will likely replace with something better. The more important thing would be to make `toolkit` more robust against this issue: https://github.com/googleapis/toolkit/issues/252

anthmgoogle:
Note: I don't have write permission on this repo and I don't have permission to see who does.

bjwatson:
@tbetbetbe Can you fix that for @anthmgoogle? I don't have admin authority.

PR #50, opened 2016-06-23
Fix artman config for core protos

shinfan:


geigerj:
LGTM

shinfan:
This should fixed https://github.com/googleapis/artman/issues/40

PR #53, opened 2016-07-12
Add per-method timeout configuration

geigerj:
Corresponds to https://github.com/googleapis/toolkit/pull/278

### DIFF
--- a/google/pubsub/v1/pubsub_gapic.yaml
+++ b/google/pubsub/v1/pubsub_gapic.yaml
@@ -60,6 +60,7 @@ interfaces:
       - topic
     retry_codes_name: idempotent
     retry_params_name: default
+    timeout_millis: 30000
     request_object_method: true
     field_name_patterns:
       name: subscription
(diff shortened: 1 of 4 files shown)


ethanbao:
LGTM

google/pubsub/v1/pubsub_gapic.yaml
```diff
@@ -60,6 +60,7 @@ interfaces:
       - topic
     retry_codes_name: idempotent
     retry_params_name: default
+    timeout_millis: 30000
```
garrettjonesgoogle: The rpc timeout in the retry configuration is 60 seconds, this should probably be the same.

geigerj: You're right, we shouldn't ever have `timeout_millis` < `initial_rpc_timeout_millis`. https://github.com/googleapis/googleapis/issues/54

Note that this change just preserved the current behavior for the Python clients and AFAIK is ignored by other languages at the moment.

PR #56, opened 2016-07-19
Add configs for cloud vision APIs.

jmuk:
- gapic.yaml is generated by toolkit
- others are hand-written

### DIFF
--- a/google/cloud/vision/v1/vision.yaml
+++ b/google/cloud/vision/v1/vision.yaml
@@ -17,10 +17,3 @@ authentication:
     - selector: '*'
       oauth:
         canonical_scopes: https://www.googleapis.com/auth/cloud-platform
-
-context:
-  rules:
-  - selector: '*'
-    requested:
-    - google.rpc.context.OriginContext
-    - google.rpc.context.ProjectContext
--- a/google/cloud/vision/v1/vision.yaml
+++ b/google/cloud/vision/v1/vision.yaml
@@ -15,7 +15,6 @@ documentation:
 authentication:
   rules:
     - selector: '*'
-      allow_without_credential: true
       oauth:
         canonical_scopes: https://www.googleapis.com/auth/cloud-platform
 
--- a/google/cloud/vision/v1/vision.yaml
+++ b/google/cloud/vision/v1/vision.yaml
@@ -1,0 +1,27 @@
+type: google.api.Service
+config_version: 2
+name: vision.googleapis.com
+title: Google Cloud Vision API
+
+apis:
+- name: google.cloud.vision.v1.ImageAnnotator
+
+documentation:
+  summary:
+    'Integrates Google Vision features, including image labeling, face, logo,
+    and landmark detection, optical character recognition (OCR), and detection
+    of explicit content, into applications.'
+
+authentication:
+  rules:
+    - selector: '*'
+      allow_without_credential: true
+      oauth:
+        canonical_scopes: https://www.googleapis.com/auth/cloud-platform
+
+context:
+  rules:
+  - selector: '*'
+    requested:
+    - google.rpc.context.OriginContext
+    - google.rpc.context.ProjectContext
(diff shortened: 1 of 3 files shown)


google/cloud/vision/v1/vision_gapic.yaml
```diff
@@ -0,0 +1,46 @@
...
+    flattening:
+      groups:
+      - parameters:
+        - requests
+    required_fields:
+    - requests
+    request_object_method: false
+    retry_codes_name: non_idempotent
+    retry_params_name: default
+    timeout_millis: 30000
```
garrettjonesgoogle: make 60000 (do we need to update the config generator?)

jmuk: Fixed manually. The issue for config generator seems filed as https://github.com/googleapis/toolkit/issues/263

garrettjonesgoogle:
LGTM

google/cloud/vision/v1/vision.yaml
```diff
@@ -0,0 +1,27 @@
...
+documentation:
+  summary:
+    'Integrates Google Vision features, including image labeling, face, logo,
+    and landmark detection, optical character recognition (OCR), and detection
+    of explicit content, into applications.'
+
+authentication:
+  rules:
+    - selector: '*'
+      allow_without_credential: true
```
bjwatson: Logging and pubsub don't have `allow_without_credential`, so it might not be needed.

garrettjonesgoogle: But wait, maybe this says something interesting - does it mean someone can call the API with empty credentials?

google/cloud/vision/v1/vision_gapic.yaml
```diff
@@ -0,0 +1,46 @@
...
+  methods:
+  - name: BatchAnnotateImages
+    flattening:
+      groups:
+      - parameters:
+        - requests
+    required_fields:
+    - requests
+    request_object_method: false
+    retry_codes_name: non_idempotent
```
bjwatson: Seems like this could be idempotent

google/cloud/vision/v1/vision.yaml
```diff
@@ -0,0 +1,27 @@
...
+      oauth:
+        canonical_scopes: https://www.googleapis.com/auth/cloud-platform
+
+context:
```
bjwatson: As @jmuk and I discussed offline, we don't seem to need this section (pubsub and logging don't have it).

google/cloud/vision/v1/vision.yaml
jmuk: Hm... Removed.

I actually brought this line from internal, but at least cloud-vision API site says about the authentication (https://cloud.google.com/vision/docs/auth-template/cloud-api-auth), so this is probably not the case of this API at least for now.

But I think gapic-generated code does not consider APIs which don't require credentials, so this is an interesting topic.

google/cloud/vision/v1/vision_gapic.yaml
jmuk: fixed

google/cloud/vision/v1/vision.yaml
jmuk: removed

google/cloud/vision/v1/vision.yaml
bjwatson: It seems that way: https://github.com/googleapis/googleapis/blob/master/google/api/auth.proto#L70

Do you think there's something we could do with this information in our generators?

bjwatson:
LGTM

google/cloud/vision/v1/vision.yaml
jmuk: yeah, generated code assumes to pass list of scopes by default, and gax-python (and other gax libraries I think) will create a channel-credential based on the scopes, which could be unnecessary.

bjwatson: IIUC this is an opportunity for efficiency improvements, but does not affect functionality, right?

PR #57, opened 2016-07-20
Remove files that don't appear in the repo. Add src-gen directory.

bjwatson:


geigerj:
This affects the manual push process from the internal source; the removed .gitignore files reflect stuff that we don't want sync'd to GitHub. We'll want to make this change when the sync process is automated, but I'd prefer to leave them in for now.

bjwatson:
Oh, got it.

PR #60, opened 2016-07-26
Remove Go-specific logic from Makefile

pongad:
Since Go is no longer generated from this repo,
make will now error if we set LANGUAGE=go.


saicheems:
LGTM

I wonder if we could add an echo that tells the user where to look if `go` is passed to the `LANGUAGE` flag.

saicheems:
Small nit: your commit title is too long

pongad:
@saicheems PTAL

saicheems:
Thanks! LGTM

PR #61, opened 2016-07-28
Add go_package attribute to logging protos

pongad:
The attribute is essentially required so that the generated client
can find what it needs

Fixes googleapis/toolkit#329.


saicheems:
LGTM

PR #55, opened 2016-07-14
Packman fix

geigerj:
This enables artman to run gRPC packman without the isGoogleApi flag.

PR #63, opened 2016-08-03
Add 'google-' prefix to cloud language API's name.

jmuk:


jmuk:
Oh, this is not a typo actually...

I've noticed this 'google-' prefix is missing as I compared with artman_logging.yaml, but actually this prefix was added by @geigerj reccently as https://github.com/googleapis/googleapis/pull/55. This is due to his recent refactoring of packman.

Either way I believe this should exist for the consistency.

garrettjonesgoogle:
Oh ok, makes sense. LGTM.

PR #64, opened 2016-08-03
Cleanup of pubsub_gapic.yaml

garrettjonesgoogle:
* Switching non-retry timeout to 60 seconds for consistency with the
  retry timeout
* Re-ordering and reformatting to reduce the edit distance with the
  config generator

### DIFF
--- a/google/pubsub/v1/pubsub_gapic.yaml
+++ b/google/pubsub/v1/pubsub_gapic.yaml
@@ -28,7 +28,7 @@ interfaces:
     - UNAVAILABLE
     - DEADLINE_EXCEEDED
   - name: non_idempotent
-    retry_codes:
+    retry_codes: []
   retry_params_def:
   - name: default
     initial_retry_delay_millis: 100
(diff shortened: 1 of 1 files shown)


google/pubsub/v1/pubsub_gapic.yaml
```diff
@@ -249,68 +249,71 @@ interfaces:
     field_name_patterns:
       topic: topic
     sample_code_init_fields:
-      - messages[0].data
+    - messages[0].data
```
michaelbausor: We could add a default initial value here. Or could do it in a followup PR

garrettjonesgoogle: I vote for following PR. I'm doing some cleanup in preparation for mixin support.

michaelbausor: OK. in that case, was the change from 30000 to 60000 for timeout intentional?

michaelbausor: Oops just saw your comment, nevermind

michaelbausor:
LGTM

PR #65, opened 2016-08-03
Move service config files to unversioned directory

geigerj:


bjwatson:
Why not do the same thing with the GAPIC config? Is it more tightly tied to the proto files?

bjwatson:
Yeah, I answered my own question. The service configs don't call out individual methods like the GAPIC config does.

LGTM

PR #67, opened 2016-08-04
Properly setup PubSub for Go

pongad:
This commit updates PubSub's go_package and GAPIC package attributes.

### DIFF
--- a/google/pubsub/v1/pubsub_gapic.yaml
+++ b/google/pubsub/v1/pubsub_gapic.yaml
@@ -6,7 +6,7 @@ language_settings:
   python:
     package_name: google.pubsub.v1
   go:
-    package_name: google.golang.org/cloud/pubsub
+    package_name: cloud.google.com/go/pubsub/apiv1
   csharp:
     package_name: Google.Pubsub.V1
   ruby:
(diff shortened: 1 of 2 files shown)


saicheems:
lgtm

google/pubsub/v1/pubsub_gapic.yaml
```diff
@@ -6,7 +6,7 @@ language_settings:
   python:
     package_name: google.pubsub.v1
   go:
-    package_name: google.golang.org/cloud/pubsub
+    package_name: cloud.google.com/go/pubsub/apiv1
```
vchudnov-g: Just checking: should this really be "apiv1" or should that last path element be split up or deleted altogether? 

FWIW, https://cloud.google.com/go/pubsub redirects to https://godoc.org/cloud.google.com/go/pubsub, and Jonathan's email suggests simply replacing s|google.golang.org/cloud/|cloud.google.com/go/| in import paths.

pongad: @vchudnov-g I just got confirmation from Dave Day. At least for now, `apiv1` -> VKit and `v1` -> Veneer, so that both can exist without stepping on the other's toes.

PR #72, opened 2016-08-05
Modify final_repo_dir for Ruby and NodeJS

jmuk:
They are going to be a part of gcloud, therefore those paths.

It's unclear what is the package names for other APIs, so their
artman configs are unchanged.


geigerj:
LGTM

Issue #77, opened 2016-08-09
cloud/runtimeconfig/v1beta1: missing dependencies


okdave:
The "runtimeconfig.proto" is missing its declared dependency "resources.proto".

This is similar to #74 in which appengine is missing its dependencies.

/cc @geigerj PTAL

geigerj:
@shinfan, please look at this one, too.

PR #82, opened 2016-08-10
Publish resources.proto

shinfan:
This is just a sync from internal repo.

Issue to fix: https://github.com/googleapis/googleapis/issues/77


geigerj:
Per offline chat, no PR is needed for syncs. Just push directly.

PR #83, opened 2016-08-11
removing type section from artman_logging

garrettjonesgoogle:
The protos in logging/type are now included in artman_core.


garrettjonesgoogle:
(see https://github.com/googleapis/googleapis/blob/master/gapic/core/artman_core.yaml for reference )

geigerj:
LGTM

PR #86, opened 2016-08-12
logging GAPIC config fixes

garrettjonesgoogle:


google/logging/v2/logging_gapic.yaml
```diff
@@ -19,7 +19,7 @@ interfaces:
 - name: google.logging.v2.ConfigServiceV2
   collections:
   - name_pattern: projects/{project}
-    entity_name: project
```
michaelbausor: Why is parent a better name that project?

garrettjonesgoogle: Because that's how the concept is referenced in the API; the parameter name in multiple functions is "parent", so we should have a corresponding "formatParentName" function (whose name is derived from this setting).

michaelbausor:
LGTM

PR #94, opened 2016-08-19
Update artmam/gapic yamls to use ${ROOTREPO} and ${TOOLKIT_HOME}

ethanbao:
Update artmam/gapic yamls to use ${ROOTREPO} and ${TOOLKIT_HOME} instead of ${THISDIR}.

This will make it possible for remote pipeline to create a isloated/temporary directory to host input/output/intermediate data/files. The downside is that it requires user to specify googleapis directory location relative to ${ROOTREPO}.

### DIFF
--- a/gapic/api/artman_cloudtrace.yaml
+++ b/gapic/api/artman_cloudtrace.yaml
@@ -1,13 +1,13 @@
 common:
   api_name: google-devtools-cloudtrace-v1
   import_proto_path:
-    - ${THISDIR}/../..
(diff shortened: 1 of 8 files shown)


michaelbausor:
LGTM

bjwatson:
Are the changes to `gapic/core/artman_core.yaml` and `gapic/lang/common.yaml` from https://github.com/ethanbao/googleapis/commit/e70ef28e7c0bb193a98d9fcbe09064539d676f50 no longer needed?

PR #96, opened 2016-08-19
Add changes for batch pipeline

michaelbausor:
Adds enable_batch_generation flags to pubsub and logging


garrettjonesgoogle:
Does this need to get updated & submitted?

michaelbausor:
Updated, PTAL

gapic/api/artman_logging.yaml
```diff
@@ -9,8 +9,10 @@ common:
   output_dir: ${REPOROOT}/artman/output
   gapic_api_yaml:
     - ${REPOROOT}/googleapis/google/logging/v2/logging_gapic.yaml
+  enable_batch_generation: True
 java:
   final_repo_dir: ${REPOROOT}/gcloud-java/gcloud-java-logging
+  batch_dir_name: gcloud-java/gcloud-java-logging
```
garrettjonesgoogle: It doesn't seem like it's necessary to configure this. I think we have enough information to make a useful path - maybe something like using ${language}/${api_name}.

gapic/api/artman_logging.yaml
```diff
@@ -11,8 +11,14 @@ common:
...
   gapic_api_yaml:
     - ${REPOROOT}/googleapis/google/logging/v2/logging_gapic.yaml
+  enable_batch_generation: True
 java:
+<<<<<<< 35a24cd143e2c6066876593c4e93e34e38449db4
   final_repo_dir: ${REPOROOT}/google-cloud-java/google-cloud-logging
+=======
+  final_repo_dir: ${REPOROOT}/gcloud-java/gcloud-java-logging
+  batch_dir_name: gcloud-java/gcloud-java-logging
+>>>>>>> Updates for batch pipeline
```
garrettjonesgoogle: ack you pushed a merge conflict

gapic/api/artman_logging.yaml
michaelbausor: It was doing that previously, I added this setting so that the output could mirror the appropriate google-cloud repository.

garrettjonesgoogle: But that seems redundant with final_repo_dir... Maybe you could break apart the field?

* final_repo_name (maybe move this to the language yaml instead of having it for every api x language)
* repo_subdir_name

And then repo_subdir_name is used by the batch pipeline, and both are used for whatever task copies the code to the final repo.

garrettjonesgoogle: Addendum - it occurs to me that final_repo_name would only apply to GCP. Maybe we need a language x org config? ("org" meaning GCP vs others)

michaelbausor: Hmmm. Now that you bring up those points and I think more about it, I am starting to think maybe the staging repo **shouldn't** mirror the google cloud repos. Adding the mirroring seems like it introduces a lot of new settings once we start thinking about non-GCP cases, staging EAP cases, and the existing single generation case, and it also ties the staging repo to the google cloud repos in an annoying way. Case in point - gcloud-java just changed to google-cloud-java. I think we may be better off having the staging repo just use `<lang>/<api>` variables like it was doing, and introduce a setting to point to the relevant google cloud repo when we add a publish step that pulls from the staging repo.

garrettjonesgoogle: Great, I like that approach.

michaelbausor:
Updated, PTAL

gapic/batch/staging.yaml
```diff
@@ -0,0 +1,6 @@
+common:
+  batch_apis: '*'
+  batch_languages: '*'
+  api_config_pattern: ${THISDIR}/../api/artman_*.yaml
+  artman_language_yaml: ${THISDIR}/../lang/common.yaml
+  batch_output_dir: ${REPOROOT}/client-library-staging/generated
```
garrettjonesgoogle: How about 'staging_output_dir' so that it can also be used when generating a single language x api? That would also require an artman option to make it put output into staging_output_dir instead of final_repo_dir. 

Sidenote, I think final_repo_dir is not terribly usable as it is set up now, and is inconsistent all over the place. So actually, I kind of like the idea of redirecting *everything* to staging_repo_dir, and leaving final_repo_dir dangling until we get it sorted out. At the same time, we could nuke the output_dir config option, and just fix it to one thing.

michaelbausor: I didn't want to include staging in the name, because of the possibility of setting it to a different and possible non-staging location

garrettjonesgoogle: I don't really think batch output should go to a different location from single language x api execution. Maybe output_dir should just be re-appropriated, and defaulted to the staging repo?

michaelbausor: So currently the batch pipeline does output everything into output_dir, just like the single language x api case. Then batch_output_dir is for an additional copy step, like the final_repo_dir in the single case. For the staging.yaml file, it is set to the staging repo, but we could have other configs like a staging_eap batch config that would point it somewhere else.

Maybe it is worth just calling it final_repo_dir in the batch config instead of batch_output_dir? But it isn't quite the same thing because final_repo_dir is actually the final directory, while batch_output_dir is extended with `<lang>/<api>` before the copy step, and no packman packaging tasks are run.

michaelbausor:
As per discussion, updated alongside https://github.com/googleapis/artman/pull/87

gapic/batch/staging.yaml
```diff
@@ -0,0 +1,5 @@
+common:
```
garrettjonesgoogle: This file is no longer staging-specific. Maybe one of these options for the filename:

* gapic/core/batch.yaml
* gapic/batch/common.yaml

michaelbausor: Good point. I prefer batch/common.yaml

michaelbausor:
PTAL

gapic/batch/common.yaml
```diff
@@ -0,0 +1,5 @@
+common:
+  batch_apis: '*'
+  batch_languages: '*'
+  api_config_pattern: ${THISDIR}/../api/artman_*.yaml
+  artman_language_yaml: ${THISDIR}/../lang/common.yaml
```
garrettjonesgoogle: Sorry, one last thing - use ${REPOROOT}/googleapis/gapic/api/* instead of ${THISDIR}/../api/* (likewise for the /lang/ case). We have removed the relative paths elsewhere.

michaelbausor: I thought we only wanted to remove them from configs that are consumed by things running in tasks, e.g. toolkit, and not configs that are processed by artman while constructing pipelines. E.g. there are still relative paths in https://github.com/googleapis/googleapis/blob/master/gapic/lang/common.yaml
Should I change them as well?

garrettjonesgoogle: I can't remember the full justification now. In any case, it seems simpler to have everything consistent. It's up to you if you want to change gapic/lang/common.yaml, but I at least wanted to apply the rule to new files.

garrettjonesgoogle:
LGTM

PR #97, opened 2016-08-23
Pubsub: Using IAM as a proper mixin

garrettjonesgoogle:
Depends on: https://github.com/googleapis/toolkit/pull/408 (Java, PHP support)

Note: This PR can't be merged until the remaining five languages are supported too.


garrettjonesgoogle:
This configuration won't work because it will mess up grpc generation. Closing now.

PR #100, opened 2016-08-23
Adding IAM protos to src_proto_path for Pub/Sub

garrettjonesgoogle:
This change will only affect grpc generation (the IAM types will be included). It won't affect GAPIC generation because GAPIC now uses the API list as the authority of interfaces to use.


michaelbausor:
LGTM

PR #105, opened 2016-08-25
Remove longrunning

geigerj:
Longrunning is already contained in googleapis-common-protos (i.e.,
the core protos), and just generating packages with artman without
special consideration results in shadowing of that package in
languages like Python.


garrettjonesgoogle:
LGTM

Issue #103, opened 2016-08-24
"Update" Cloud Speech to v1beta1


bjwatson:
### What:
Update Cloud Speech: [v1](https://github.com/googleapis/googleapis/tree/master/google/cloud/speech/v1) -> [v1beta1](https://github.com/googleapis/googleapis/tree/master/google/cloud/speech/v1beta1)

### Why:
The Cloud Speech team mistakenly provided v1 protos. The correct version for now is v1beta1.

bjwatson:
FYI @garrettjonesgoogle

PR #106, opened 2016-08-26
Switch back to Cloud Speech v1beta1.

bjwatson:
Although v1 sounds newer, that was a mistake. v1beta1 is the version in
active development right now.

Fixes https://github.com/googleapis/googleapis/issues/103


bjwatson:
FYI @anthmgoogle, this depends on our LRO story.

bjwatson:
FYI @tseaver, @dhermes, @daspecster... good thing you hadn't started working on Python Speech, yet. We just learned this week that we had been building from the wrong version. We'll publish updated gRPC and GAX packages as soon as we nail this down.

bjwatson:
@jmuk @garrettjonesgoogle Can you check what surface Ruby, Node, and PHP GCloud are building against for Speech? I suspect it's out of date.

@jskeet @chrisdunelm I don't know if you started on this for .NET yet, but the surface is a bit different now.

FYI @omaray

jmuk:
Ugh, v1 is merged for Ruby (https://github.com/GoogleCloudPlatform/google-cloud-ruby/tree/master/google-cloud-speech/lib/google/cloud/speech/v1). Node is not yet merged.

If v1beta1 is the newer one (and the one actively developed), then... what should I do?  Remove the existing v1 and re-generate v1beta1?  cc: @blowmage @quartzmo

blowmage:
Yeah, remove v1 and add the right version to the ruby repo.

bjwatson:
@jmuk Can you help with finishing the GAPIC config? Attached is a patch file showing hand-edits we had done to the v1 GAPIC config, for which we might still need to do something analogous to the v1beta1 config.

We also need to figure out if we want to represent LRO the way it appears in this auto-gen GAPIC config.

google/cloud/speech/cloud_speech.yaml
```diff
@@ -11,17 +11,18 @@ documentation:
     Google Cloud Speech API.
 
 apis:
-- name: google.cloud.speech.v1.Speech
+- name: google.cloud.speech.v1beta1.Speech
+- name: google.longrunning.Operations
```
garrettjonesgoogle: This is going to create a problem in the scenario that a client uses multiple APIs that have long-running operations; the long-running operations class(es) will be duplicated and potentially clash.

jmuk: Agreed. It is still unclear to us about the long running operations.
Can we exclude AsyncRecognize and longrunning.Operations for the time being?

google/cloud/speech/v1beta1/cloud_speech_gapic.yaml
```diff
@@ -0,0 +1,156 @@
...
+    flattening:
+      groups:
+      - parameters:
+        - config
+        - audio
+    required_fields:
+    - config
+    - audio
+    request_object_method: true
+    retry_codes_name: non_idempotent
```
jmuk: I think SyncRecognize would be an `idempotent` operation (the configgen estimates the retry_codes_name from the HTTP method name in the config, that is a good guess but sometimes set wrongly).

google/cloud/speech/v1beta1/cloud_speech_gapic.yaml
```diff
@@ -0,0 +1,156 @@
...
```
jmuk: Similarly, idempotent?

google/cloud/speech/v1beta1/cloud_speech_gapic.yaml
```diff
@@ -0,0 +1,156 @@
...
+    timeout_millis: 60000
+  - name: StreamingRecognize
+    flattening:
+      groups:
+      - parameters:
+        - streaming_config
+        - audio_content
+    required_fields:
+    - streaming_config
+    - audio_content
```
jmuk: Remove the flattening and required_fields configs, we will not use them for bidi-streams probably.

google/cloud/speech/cloud_speech.yaml
garrettjonesgoogle: I think we should do that for now.

bjwatson: Agreed. How do we want to do this? @pongad did a `toolkit` change to ignore streaming methods; shall we do the same for LRO methods? Alternatively, we can comment it out in the proto and configs, but that's messier. (I prefer commenting out vs. deleting in this case, because it makes it clear that we didn't forget to include it, and we can associate a TODO comment explaining under what conditions we'll un-comment it)

garrettjonesgoogle: I recommend against commenting it out in the proto. That makes integration from the source of truth harder. I also recommend against commenting it out in config, in the transitional case where some languages support it but not others. I think the best approach is through ignoring the methods in code, so each language is capable of including them when they support LRO methods.

bjwatson: SGTM. @pongad do you have a little time on your Monday to repeat the same trick with LRO that you did with streaming?

bjwatson:
I addressed all PR comments, except for the LRO issue.

garrettjonesgoogle:
I want to regenerate Speech for Java today - is someone handling the LRO issue or should I take that up?

jmuk:
It turns out that commenting out LRO method is not a good solution -- we should keep providing LRO methods at least in gRPC level, so that hand-written layer can choose to use it with manual effort.

Probably we should
- remove `google.longrunning.Operations` service from the config
- but keep the method itself. So the method itself is visible from gRPC level and Gapic. However no Gapic code to handle the return value
- revisit later how the Gapic or GAX should handle the LRO value

bjwatson:
@garrettjonesgoogle I agree with @jmuk's suggestion to remove `google.longrunning.Operations`, but keep the `AsyncRecognize` method that returns an LRO object. The latter is an important method for the google-cloud teams.

We're working on fixing some issues in the remote pipeline (Zookeeper is being unresponsive), so can you re-generate the GAPIC config if you have a working version of the local pipeline? (I currently don't)

anthmgoogle:
Agree that we should leave the LRO method in for the short term. We need to determine per/language if we need to special case LRO for beta. But I would not block pushing wrappers on this, and it is less likely to be to need to be different in the short term for dynamic languages.

bjwatson:
@jmuk I merged in updates from @garrettjonesgoogle implementing your suggestions. I think this is ready to merge. PTAL.

jmuk:
LGTM

jskeet:
Thanks for the heads-up. Will delete the V1 code from the google-cloud-dotnet repo, but will probably wait until Chris is back from holiday before regenerating. (At that point, I should definitely make sure I know how to regenerate it myself...)

jskeet:
Could I suggest that if https://github.com/googleapis/googleapis/tree/master/google/cloud/speech/v1 really isn't ready, we add a readme or something indicating that? Or just delete the directory?

bjwatson:
Good point @jskeet. I removed that directory. @jmuk and @garrettjonesgoogle, PTAL

garrettjonesgoogle:
ship it

bjwatson:
Done

PR #108, opened 2016-08-26
Add non-admin bigtable configs

geigerj:
Includes service, artman, GAPIC configs with hand-edits to required fields in the GAPIC config.

Note that this does not address the Bigtable admin APIs.

### DIFF
--- a/google/bigtable/bigtable.yaml
+++ b/google/bigtable/bigtable.yaml
@@ -1,0 +1,34 @@
+# Google Bigtable API service configuration
+
+type: google.api.Service
+config_version: 0
+name: bigtable.googleapis.com
+
+title: Google Cloud Bigtable API
+
+documentation:
+  summary:
+    Google Cloud Bigtable - http://cloud.google.com/bigtable/
+
+apis:
+- name: google.bigtable.v2.Bigtable
+
+authentication:
+  rules:
+    # Unless explicitly weakened, all ops require write access
+    - selector: '*'
+      oauth:
+        canonical_scopes: https://www.googleapis.com/auth/bigtable.data,
+                          https://www.googleapis.com/auth/cloud-bigtable.data,
+                          https://www.googleapis.com/auth/cloud-platform
+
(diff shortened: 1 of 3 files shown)


google/bigtable/bigtable.yaml
```diff
@@ -0,0 +1,34 @@
+# Google Bigtable API service configuration
+
+type: google.api.Service
+config_version: 0
```
garrettjonesgoogle: I just noticed that we're using different config versions all over; pubsub is using version 1, logging is using version 2, and you have version 0 here. Should we settle on one version? (We have only one version of code in toolkit reading these files)

google/bigtable/v2/bigtable_gapic.yaml
```diff
@@ -0,0 +1,131 @@
+type: com.google.api.codegen.ConfigProto
+generate_samples: true
```
garrettjonesgoogle: sidenote, unrelated to this review - we should get rid of this setting, it doesn't actually impact anything any more.

google/bigtable/bigtable.yaml
geigerj: These are derived from the Google-internal source of truth; I think they have effects in API compiler but am not positive.

garrettjonesgoogle: Ok. We can leave them as-is.

garrettjonesgoogle:
LGTM

google/bigtable/bigtable.yaml
bjwatson: They're also a way of tracking if we have the latest version from Google.

google/bigtable/v2/bigtable_gapic.yaml
bjwatson: ACK

bjwatson: See https://github.com/googleapis/googleapis/pull/111

PR #110, opened 2016-08-26
Adding IAM method config for Pub/Sub

garrettjonesgoogle:
This will only take effect for languages which have implemented reroute_to_grpc_interface.

### DIFF
--- a/google/pubsub/v1/pubsub_gapic.yaml
+++ b/google/pubsub/v1/pubsub_gapic.yaml
@@ -176,6 +176,52 @@ interfaces:
     field_name_patterns:
       subscription: subscription
     timeout_millis: 60000
+  - name: SetIamPolicy
+    flattening:
+      groups:
+      - parameters:
+        - resource
+        - policy
+    required_fields:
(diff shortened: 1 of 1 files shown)


michaelbausor:
If a language hasn't implemented the reroute_to_grpc_interface, won't they start generating non-remapped IAM methods that won't work?

garrettjonesgoogle:
No, the existing languages use the list of methods in the Interface, not the list of configured methods.

michaelbausor:
So for those languages, it is not an error to have methods in the config that don't correspond to any methods in the protos? And while they get the methods from the interface not the config, they get the list of interfaces from the config, not the protos? Otherwise they would not like the addition of the IAM proto in the artman file.

LGTM, I just didn't realize this was how it worked.

garrettjonesgoogle:
That is correct, it is not an error to have non-existent methods configured. 

For all languages, they don't get the list of interfaces from the GAPIC config either, they get that from the service yaml (under the key 'apis').

PR #113, opened 2016-08-29
Fix Ruby configurations.

jmuk:
- Add 'Cloud' to Ruby packages.
  The decision is that gapic files will be under cloud, so
  change the package name to do so.
- Replace 'gcloud-ruby' by 'google-cloud-ruby'
  The repository name change happened quite recently.


jmuk:
@Landrito, can you take a look?

Landrito:
I think you are missing Bigtable.

jmuk:
Added bigtable.

Landrito:
LGTM

PR #123, opened 2016-09-07
Add method flattenings

chrisdunelm:
For RunQuery and Commit


jmuk:
LGTM. Thanks for catching.
It seems the required_parameters don't look great as well -- I'll fix them on my side.

PR #124, opened 2016-09-07
Fix required_parameters of datastore APIs

jmuk:
- RunQuery -- "query" and "gql_query" are oneof. They shouldn't
  appear at the same time as the required parameters.
- Commit -- The comment says "transaction" is optional.

Related: https://github.com/googleapis/googleapis/pull/123


chrisdunelm:
LGTM

PR #130, opened 2016-09-09
Fix some paths to be relative to REPOROOT

geigerj:
Previously they were relative to THISDIR, which is not safe when
running remotely.


shinfan:
LGTM

bjwatson:
@geigerj @ethanbao Can we purge THISDIR from artman altogether, or is it still needed for some use cases?

geigerj:
I removed the remaining macros, PTAL.

bjwatson:
LGTM

PR #133, opened 2016-09-12
Use uppercamel for ruby package name for Errorreporting

swcloud:


jmuk:
LGTM

bjwatson:
@swcloud Can you do something similar for Python? (i.e. `errorreporting` => `error_reporting`)

@jmuk @garrettjonesgoogle @shinfan @jskeet @chrisdunelm @vchudnov-g @pongad Do you want something similar for your languages? I suspect this was just the behavior of the GAPIC config generator, and not a conscious choice on anyone's part.

jskeet:
@bjwatson: Unfortunately there's more than just gapic involved here - there's protoc and the gRPC plugin. Those will both take the namespace from the proto file - it would be odd to have the `Google.Cloud.ErrorReporting.V1` package contain a `Google.Cloud.Errorreporting.V1` namespace.

Fundamentally, I think we need to take more care over naming very early on in the API process, and make sure that it will be appropriate for all platforms.

jmuk:
@jskeet, I believe this is mostly for Gapic namespace only. gRPC namespace for tracing is actually `google.devtools.clouderrorreporting.v1` and that's not going to change, while Gapic can choose a slightly different name space (we do that for pubsub already -- gRPC is `google.pubsub.v1` while Gapic is under `Google::Cloud::Pubsub::V1` in Ruby).

Having said that, apparently this will be highly language dependent. Rubyists seem to prefer camelcase for the name space, so it's great we could do that and that's why this PR was created. If people in the language community don't care much, I think one-word 'errorreporting' (and 'Errorreporting' when capitalized) is just fine.

For Node.JS, it could be `@google-cloud/error-reporting`, but I am not sure how many people would care. Until someone raises voice on this topic, I'd like to keep this as-is for now.

jskeet:
I'm pretty sure people would care in the C# community. ErrorReporting feels significantly more pleasant than Errorreporting to me.

I'm bringing it up internally for API production.

PR #134, opened 2016-09-13
Move Python autogen packages to google.cloud.gapic namespace

bjwatson:
Avoid conflicts with hand-written google-cloud-python code.

Updates https://github.com/GoogleCloudPlatform/google-cloud-python/issues/2258

### DIFF
--- a/google/cloud/language/v1beta1/language_gapic.yaml
+++ b/google/cloud/language/v1beta1/language_gapic.yaml
@@ -3,7 +3,7 @@ language_settings:
   java:
     package_name: com.google.cloud.language.spi.v1beta1
   python:
-    package_name: google.cloud.gax.language.v1beta1
+    package_name: google.cloud.gapic.language.v1beta1
   go:
     package_name: cloud.google.com/go/language/apiv1beta1
   csharp:
--- a/google/cloud/language/v1beta1/language_gapic.yaml
+++ b/google/cloud/language/v1beta1/language_gapic.yaml
@@ -3,7 +3,7 @@ language_settings:
   java:
     package_name: com.google.cloud.language.spi.v1beta1
   python:
-    package_name: google.cloud.language.v1beta1
+    package_name: google.cloud.gax.language.v1beta1
   go:
     package_name: cloud.google.com/go/language/apiv1beta1
   csharp:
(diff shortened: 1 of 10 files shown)


google/cloud/language/v1beta1/language_gapic.yaml
```diff
@@ -3,7 +3,7 @@ language_settings:
   java:
     package_name: com.google.cloud.language.spi.v1beta1
   python:
-    package_name: google.cloud.language.v1beta1
```
garrettjonesgoogle: Why are you putting the generated packages under the GAX namespace? GAX = Google Api eXtensions, which is supposed to be exclusively the API-agnostic run-time library.

bjwatson: https://github.com/GoogleCloudPlatform/google-cloud-python/issues/2258

Basically, an element in a Python path can be a "namespace" or it can define "members", but not both. Our current scheme of putting hand-written code in `google.cloud.{api}` and auto-gen code in `google.cloud.{api}.{version}` means that `google.cloud.{api}` is a namespace, which interferes with the google-cloud-python's team use of members to improve developer experience.

I can forward you an email with the full discussion, and you can also take a look at the Python section of the CCL Namespacing PRD to see the schema we're moving toward.

bjwatson: Also, we call https://github.com/googleapis/gax-python [`google-gax`](https://pypi.python.org/pypi/google-gax) in Python-land.

bjwatson: We discussed offline and decided to use `gapic` rather than `gax`. The former means Google API Codegen, and is a more appropriate name than the latter, which mean Google API eXtensions (which specifically refers to [gax-python](https://github.com/googleapis/gax-python)).

Also, we will soon change the Python package names on PyPI from `gax-google-*` to `gapic-google-*` (not part of this PR). It's much easier to make this change now than it is post-1.0.

FYI @tseaver @dhermes @daspecster @jgeewax

bjwatson:
@geigerj @garrettjonesgoogle PTAL.

We want this change for the admin APIs we're currently generating in MTV. FYI @ethanbao @swcloud @Landrito @jcanizales

geigerj:
LGTM

Is there an issue tracking the corresponding work in `packman` or `artman` to make this happen for gRPC packages as well?

garrettjonesgoogle:
LGTM

bjwatson:
@geigerj It does now: https://github.com/googleapis/artman/issues/97

PR #135, opened 2016-09-14
Skip packman for NodeJS/Ruby logging

geigerj:
Similar flags should be put in for other APIs that also require this.


jmuk:
LGTM. Who's gonna do for other APIs?

geigerj:
@jmuk Could you add this flag to the other APIs since you have some context on the state of [GoogleCloudPlatform/google-cloud-ruby](https://github.com/GoogleCloudPlatform/google-cloud-ruby)?

PR #145, opened 2016-09-20
CloudDebugger v2 configs

jcanizales:


gapic/api/create_new_config.sh
```diff
@@ -0,0 +1,53 @@
...
+# ./create_new_config.sh clouddebugger v2 devtools/
+#
+# Dependencies need to be added manually after executing.
+#
+# Args:
+# $1 API name (e.g. pubsub)
+# $2 API version (e.g. v1)
+# $3 Path within googleapis/google/ (e.g. devtools/). Without the API name.
+
+API_NAME=$1
```
garrettjonesgoogle: All generation should be driven by artman. We shouldn't have any scripts in googleapis.

google/devtools/clouddebugger/clouddebugger.yaml
```diff
@@ -0,0 +1,38 @@
+type: google.api.Service
+
+config_version: 2 # (assumed)
```
garrettjonesgoogle: assumed? from what?

jcanizales:
For the script, I don't know if we have some standard way to declare input parameters, as in python.

google/devtools/clouddebugger/clouddebugger.yaml
geigerj: 2 is correct -- check internal source of truth

gapic/api/create_new_config.sh
jcanizales: I'm confused - doesn't artman needs this config file to run?

google/devtools/clouddebugger/clouddebugger.yaml
jcanizales: Assumed by the script in https://github.com/googleapis/googleapis/pull/146 when it inputs a service config without a version set.

gapic/api/create_new_config.sh
garrettjonesgoogle: Yes, but artman is responsible for gapic config generation and surface code generation; if we are going to generate artman configuration, artman should do that too.

google/devtools/clouddebugger/clouddebugger.yaml
geigerj: We shouldn't make this assumption. The config version is available from the original internal service config.

gapic/api/create_new_config.sh
jcanizales: What's involved in doing that? There are already 11 configs created manually in the repo, I bet by the same procedure in this script, just executed by a human.

geigerj: There's an example in the artman repo of a simple pipeline that doesn't require a config as input: https://github.com/googleapis/artman/blob/master/USAGE.rst#local

If you want, we can talk offline about how to integrate this with artman.

google/devtools/clouddebugger/clouddebugger.yaml
```diff
@@ -0,0 +1,73 @@
...
+    get: /v2/debugger/debuggees/{debuggee_id}/breakpoints
+
+  - selector: google.devtools.clouddebugger.v2.Debugger2.ListDebuggees
+    get: /v2/debugger/debuggees
+
+  - selector: google.devtools.clouddebugger.v2.Controller2.RegisterDebuggee
+    post: /v2/controller/debuggees/register
+    body: '*'
+  
+
```
geigerj: nit: whitespace? (here and throughout this file)

gapic/api/artman_clouddebugger.yaml
```diff
@@ -0,0 +1,30 @@
...
+  final_repo_dir: ${REPOROOT}/artman/output/gcloud-python-clouddebugger
+go:
+  final_repo_dir: ${REPOROOT}/gapi-clouddebugger-go
+csharp:
+  final_repo_dir: ${REPOROOT}/artman/output/gcloud-csharp-clouddebugger
+php:
+  final_repo_dir: ${REPOROOT}/artman/output/gcloud-php-clouddebugger
+ruby:
+  final_repo_dir: ${REPOROOT}/google-cloud-ruby/google-cloud-clouddebugger
+  skip_packman: True
```
geigerj: It doesn't look like google-cloud-ruby has a debugger package, so probably we don't want this flag. Please check with @jmuk. (This flag should be set if google-cloud-ruby intends to put a handwritten layer on top of this.)

Same with nodejs.

jmuk: @geigerj is right, this is not necessary.

jcanizales: What's an automatable way to tell?

jcanizales: @jmuk how do you know? :)

jmuk: There's no automated way. Check with https://github.com/GoogleCloudPlatform/google-cloud-ruby repository (and -node for nodejs).

jcanizales: That's problematic, as we can't have a VKit engineer spend time figuring that out for every API. Is there a way we can remove the need to specify it, or, at least, move the decision to a later stage?

google/devtools/clouddebugger/clouddebugger.yaml
jcanizales: Regenerated the file with the proper inputs.

google/devtools/clouddebugger/clouddebugger.yaml
jcanizales: Regenerated without the spurious whitespace.

gapic/api/artman_clouddebugger.yaml
geigerj: I don't think this is problematic.

- There is a small, finite number of existing APIs that have handwritten client libraries, for which this flag needs to be set by hand.
- The cost of creating new handwritten client libraries is very high compared to the cost of manually setting this flag.
- There will always be some configuration required for each API we generate.
- If there are values that are always deterministically knowable from the input, they should be set in the code rather than through configuration.

(That's not to say, of course, that we shouldn't make an effort to minimize the amount of configuration work necessary.)

google/devtools/clouddebugger/v2/clouddebugger_gapic.yaml
```diff
@@ -0,0 +1,167 @@
...
+    - breakpoint_id
+    - client_version
+    request_object_method: true
+    retry_codes_name: idempotent
+    retry_params_name: default
+    timeout_millis: 60000
+  - name: ListBreakpoints
+    required_fields:
+    - debuggee_id
+    - include_all_users
```
geigerj: According to the proto, this field has a default value if not set: https://github.com/googleapis/googleapis/blob/5fd17fb8a2559b0c5d06b65f8a19d9cbbeb79bce/google/devtools/clouddebugger/v2/debugger.proto#L133

Here and throughout, you need to hand-edit the required fields based on the semantics described in the proto; our tooling is currently unable to determine what's required/optional. Similarly for flattening.

gapic/api/artman_clouddebugger.yaml
jcanizales: > There will always be some configuration required for each API we generate.

Reactive corrections to problems identified in the generated library are fine. Configuration to be set (and reviewed) before the library can be generated is problematic, for being hard to scale.

> The cost of creating new handwritten client libraries is very high compared to the cost of manually setting this flag.

I'm not worried about adding a small cost to creating a new handwritten library. I'm worried about an engineer having to check whether a handwritten library exists or not before being able to autogenerate a library. We want to be able to generate libraries for hundreds of APIs (and, ideally, an unbound amount): every cost in the process needs to have that multiplier applied before considering its magnitude.

> There is a small, finite number of existing APIs that have handwritten client libraries, for which this flag needs to be set

So a machine-readable list should be easy to create? :)

But even so, I think the decision should be made at a later stage: Say Jun or a computer verifies that there's no hand-written library for this API today, and we proceed all the way with the generation. What happens when some time in the future they want to write a hand-written library on top?

google/devtools/clouddebugger/v2/clouddebugger_gapic.yaml
jcanizales: All proto3 primitive types have a default value! So this should never happen :) (We probably want to still "require" string fields, like resource ids, but meaning non-empty).

What made the tool think this field is required? The HTTP pattern for `ListBreakpoints` is `/v2/debugger/debuggees/{debuggee_id}/breakpoints`.

geigerj: Probably "default value" was the wrong choice of words. I meant that the API has a default behavior (other than returning an illegal argument error) when this field is set to the default proto3 value.

The tool doesn't look at HTTP patterns. It would be a good starting heuristic if someone wanted to invest the effort, but AFAIK the only way to tell whether a field is required with full confidence is to read the comments on the proto. @saicheems is working on a way to specify this through an annotation in the proto comments that can be read by the tooling, but it's manual for now.

gapic/api/artman_clouddebugger.yaml
geigerj: This flag shouldn't be set unless we're integrating with a handwritten library. So if we're scaling to an unbounded amount of APIs, this flag will essentially never be set because handwritten libraries don't scale to an unbounded amount.

It seems to me that keeping a machine readable list of APIs that require this flag is doing the exact same thing as setting this flag (i.e., must set one line of code per API per language with source delivery that cares about package metadata), except that it introduces yet another input in yet another location to the generation logic.

If there's a solution that lets us integrate with handwritten libraries without having to set a configuration flag, I'm all for it.

google/devtools/clouddebugger/v2/clouddebugger_gapic.yaml
jcanizales: Ok, "required === default value is an error" makes more sense. Is that actually what the generated library enforces? E.g. the user can't pass `0` for required numbers, `""` for required strings, and `[]` for required bytes?

It still shouldn't ever happen for boolean fields like this one :) It would mean "this field needs to be set to true", which doesn't make sense.

From inspecting the generated config - the tool is just assuming all fields of the request are required?

jcanizales: FWIW, an HTTP pattern that uses GET tells you with full confidence that the fields that don't appear in the pattern are optional (as HTTP GET requests can't have bodies). And pretty much full confidence that string fields that appear in the pattern are required (as in "can't be empty").

gapic/api/artman_clouddebugger.yaml
jcanizales: So in the case of this API, for which the flag isn't set, what happens if, after the generated libraries are published, they want to write a hand-written library on top?

> It seems to me that keeping a machine readable list of APIs that require this flag is doing the exact same thing as setting this flag

Not the same: Setting this flag right now requires 2 lookups and 2 reviews for every API, so multiple hundreds. Listing the APIs with hand-written libraries is writing a list with 11 + 14 entries, that gets updated rarely.
Ask yourself, what would take you less time: list the existing veneers, or check one by one each API to see if it has a veneer? Even if you did the latter, your brain would end up remembering the list of veneers to save you trouble.

google/devtools/clouddebugger/v2/clouddebugger_gapic.yaml
geigerj: > Is that actually what the generated library enforces? E.g. the user can't pass 0 for required numbers, "" for required strings, and [] for required bytes?

No, it doesn't enforce the restriction. It just makes those parameters mandatory in languages (like Python) that distinguish required from keyword/optional arguments, making it less likely for users to leave them unset accidentally. AFAIK, there's no technical reason that we couldn't validate, though.

> From inspecting the generated config - the tool is just assuming all fields of the requests are required?

Yes. As I said before, a better heuristic would be helpful.

> FWIW, an HTTP pattern that uses GET tells you with full confidence that the fields that don't appear in the pattern are optional (as HTTP GET requests can't have bodies). And pretty much full confidence that string fields that appear in the pattern are required (as in "can't be empty").

Right, unless someone is silly and decides to make a query parameter semantically required. If you're interested in improving how we handle required/optional params, please coordinate with @saicheems, since I think he's already started working on that.

gapic/api/artman_clouddebugger.yaml
geigerj: Ok, fine by me if you want to switch this flag to a separately-stored metadata file.

jcanizales: Offline conversation for the record:

* If the flag isn't set, and later they add a hand-written library on top, they just need to modify auto-generated package metadata files in the repo.
* If the flag isn't set, and a hand-written library already exists, those auto-generated files would need to be reverted manually before merging the generated files in the library's repo.

In both cases, one can go and set the flag after the fact so future runs don't recreate those files, without needing to restart the whole process.

google/devtools/clouddebugger/v2/clouddebugger_gapic.yaml
```diff
@@ -0,0 +1,167 @@
+type: com.google.api.codegen.ConfigProto
+language_settings:
+  java:
+    package_name: com.google.cloud.devtools.clouddebugger.spi.v2
+  python:
+    package_name: google.cloud.devtools.clouddebugger.v2
```
bjwatson: Recent rename: `google.cloud.gapic.devtools.clouddebugger.v2 `

@geigerj just fixed the GAPIC config generator to address this in the future.

jmuk: Wait a min, this reminds me of the error reporting package naming.

In cloud error reporting API, the grpc namespace is somehow `google.devtools.clouderrorreporting`, but the namespace for the Gapic will be different, a bit more manual edit to name them more understandable. See https://github.com/googleapis/googleapis/blob/master/google/devtools/clouderrorreporting/v1beta1/errorreporting_gapic.yaml, the python package names are more like `google.cloud.gapic.errorreporting`.

If that's so, we should do the same thing here as well, omitting `devtools` and `cloud` prefix. Python naming would be `google.cloud.gapic.debugger.v2`,  Ruby `Google::Cloud::Debugger::V2`, node `@google-cloud/debugger`, go `cloud.google.com/go/debugger/apiv2`, java `com.google.cloud.debugger.spi.v2`?  @swcloud @garrettjonesgoogle , @pongad wdyt?

bjwatson: Good catch @jmuk! The Python namespace is `google.cloud.grpc.{api}.{version}` and `google.cloud.gapic.{api}.{version}`.

garrettjonesgoogle: Yes, that is what we need for Java - remove 'cloud' and 'devtools'.

google/devtools/clouddebugger/clouddebugger.yaml
```diff
@@ -0,0 +1,70 @@
...
+documentation:
+  summary: |
+    Examines the call stack and variables of a running application
+    without stopping or slowing it down.
+
+apis:
+- name: google.devtools.clouddebugger.v2.Debugger2
+- name: google.devtools.clouddebugger.v2.Controller2
+
+http:
```
bjwatson: Remove this section

google/devtools/clouddebugger/clouddebugger.yaml
```diff
@@ -0,0 +1,70 @@
...
+        https://www.googleapis.com/auth/cloud-platform
+  - selector: |-
+      google.devtools.clouddebugger.v2.Controller2.RegisterDebuggee,
+      google.devtools.clouddebugger.v2.Controller2.ListActiveBreakpoints,
+      google.devtools.clouddebugger.v2.Controller2.UpdateActiveBreakpoint
+    oauth:
+      canonical_scopes: |-
+        https://www.googleapis.com/auth/cloud_debugger,
+        https://www.googleapis.com/auth/cloud-platform
+context:
```
bjwatson: Remove this section

google/devtools/clouddebugger/clouddebugger.yaml
```diff
@@ -0,0 +1,70 @@
...
+
+  - selector: google.devtools.clouddebugger.v2.Controller2.ListActiveBreakpoints
+    get: /v2/controller/debuggees/{debuggee_id}/breakpoints
+
+  - selector: google.devtools.clouddebugger.v2.Controller2.UpdateActiveBreakpoint
+    put: /v2/controller/debuggees/{debuggee_id}/breakpoints/{breakpoint.id}
+    body: '*'
+
+authentication:
+  rules:
```
bjwatson: The `canonical_scopes`, so can we simplify this with `- selector: '*'`?

google/devtools/clouddebugger/clouddebugger.yaml
jcanizales: Why? It's information we leverage (for idempotency, potentially for identifying required vs optional fields), and we leave it in the proto annotations when the API authors set it that way.

google/devtools/clouddebugger/clouddebugger.yaml
jcanizales: We can, but not easily in an automated way. How important is it?

google/devtools/clouddebugger/clouddebugger.yaml
jcanizales: Done.

google/devtools/clouddebugger/clouddebugger.yaml
garrettjonesgoogle: When APIs are open-sourced, all of the http information from the original service yaml is copied onto the protos. Thus, this becomes redundant.

jcanizales: 👍

google/devtools/clouddebugger/clouddebugger.yaml
bjwatson: not important

google/devtools/clouddebugger/v2/clouddebugger_gapic.yaml
```diff
@@ -0,0 +1,162 @@
+type: com.google.api.codegen.ConfigProto
+language_settings:
+  java:
+    package_name: com.google.cloud.devtools.clouddebugger.spi.v2
+  python:
+    package_name: google.cloud.gapic.devtools.clouddebugger.v2
```
bjwatson: We need to follow the namespacing guidelines in the PRD I shared with you. You can compare against other GAPIC configs to see the correct pattern.

For example, Python is `google.cloud.gapic.{api}.{version}`, so it would be `google.cloud.gapic.debugger.v2` in this case.

garrettjonesgoogle: Java should be `com.google.cloud.debugger.spi.v2`

jcanizales: To confirm, we want to change the API name (`{api}`) from `clouddebugger` to `debugger`, right?

Do we want that to happen upstream, or do we want our tools to notice the prefix in the name and strip it?

jcanizales: (That is, apart from removing the directory path `devtools` from the package name, which I'll do too).

geigerj: I had originally hoped that this could be derived from the [`title` attribute](https://github.com/googleapis/googleapis/pull/145/files/0415ff5d6d0d2ede2dfa54b117fc6341de4c7939#diff-33c86e384db847920bc6861fd089906eR6) of the service config via googleapis/toolkit#270. @garrettjonesgoogle also has a proposed introducing a "short name" attribute into configuration for deriving the output directories, since it may not be possible to automatically derive this from service config -- if that happens, I'd assume it would coincide with `{api}` here as well?

jcanizales: Is there any reason (other than backwards-compatibility) not to make this change upstream, in the protos?

geigerj: By making this change upstream, you mean changing the proto package from `google.devtools.clouddebugger` to `google.debugger.v2`, right? The main reason I'm aware of for not trying to do this is just pragmatics, but I don't actually know what sort of work/coordination would be required to make this happen...

jcanizales: Well the proto package needs to match the directory structure, so the minimum change would be to move the API from `google/devtools/clouddebugger/` to `google/devtools/debugger/`. That would give us the API name we want for it.

Then if we also want it out of `/devtools/`, or into `/cloud/`, sure; but that's not a problem really: `google.cloud.gapic.{api}.{version}` is the same no matter the subdirectory.

garrettjonesgoogle: I think the issue with changing proto paths/directories is that there are potentially other consumers outside of the vkit toolchain that depend on the protos. But, one thing to consider is that this was only just recently released, so potentially the impact would be minimal to change it; the flip side is there might be internal consumers who depend on the current structure. I think the larger issue is that we just don't know the impact and we don't have a policy around it - we aren't following something like semantic versioning for protos. 

Absent changing that, I think we just have to define the api short name ourselves.

jcanizales: Regenerated as `com.google.cloud.debugger.spi.v2` and `google.cloud.gapic.debugger.v2`.

google/devtools/clouddebugger/v2/clouddebugger_gapic.yaml
```diff
@@ -0,0 +1,159 @@
...
+    required_fields:
+    - debuggee_id
+    - breakpoint_id
+    - client_version
+    request_object_method: true
+    retry_codes_name: idempotent
+    retry_params_name: default
+    timeout_millis: 60000
+  - name: ListBreakpoints
+    required_fields:
```
jcanizales: This didn't generate a flattening section because the request has more than 4 fields. Only two of them are required, though: Should we apply the flattening threshold only to the number of required fields?

google/devtools/clouddebugger/v2/clouddebugger_gapic.yaml
```diff
@@ -0,0 +1,159 @@
...
+    request_object_method: true
+    retry_codes_name: idempotent
+    retry_params_name: default
+    timeout_millis: 60000
+  - name: ListDebuggees
+    flattening:
+      groups:
+      - parameters:
+        - project
+        - include_inactive
```
jcanizales: Should non-required parameters be in the flattening group?

google/devtools/clouddebugger/v2/clouddebugger_gapic.yaml
```diff
@@ -0,0 +1,159 @@
...
+    retry_codes_name: non_idempotent
+    retry_params_name: default
+    timeout_millis: 60000
+  - name: ListActiveBreakpoints
+    flattening:
+      groups:
+      - parameters:
+        - debuggee_id
+        - wait_token
+        - success_on_timeout
```
jcanizales: Ditto here with flattened optional parameters.

google/devtools/clouddebugger/clouddebugger.yaml
jcanizales: Done.

google/devtools/clouddebugger/v2/clouddebugger_gapic.yaml
jcanizales: Done.

google/devtools/clouddebugger/v2/clouddebugger_gapic.yaml
garrettjonesgoogle: I think it cases where the number of required fields doesn't surpass the threshold, we may want to make a flattening group for them.

gapic/api/artman_clouddebugger.yaml
```diff
@@ -0,0 +1,28 @@
...
+go:
+  final_repo_dir: ${REPOROOT}/gapi-clouddebugger-go
+csharp:
+  final_repo_dir: ${REPOROOT}/artman/output/gcloud-csharp-clouddebugger
+php:
+  final_repo_dir: ${REPOROOT}/artman/output/gcloud-php-clouddebugger
+ruby:
+  final_repo_dir: ${REPOROOT}/google-cloud-ruby/google-cloud-clouddebugger
+nodejs:
+  final_repo_dir: ${REPOROOT}/gcloud-node/packages/clouddebugger
```
garrettjonesgoogle: take 'cloud' out of all of the final repo dir paths

google/devtools/clouddebugger/v2/clouddebugger_gapic.yaml
```diff
@@ -0,0 +1,159 @@
...
+  go:
+    package_name: cloud.google.com/go/devtools/clouddebugger/apiv2
+  csharp:
+    package_name: Google.Devtools.Clouddebugger.V2
+  ruby:
+    package_name: Google::Cloud::Devtools::Clouddebugger::V2
+  php:
+    package_name: Google\Cloud\Devtools\Clouddebugger\V2
+  nodejs:
+    package_name: '@google-cloud/clouddebugger'
```
garrettjonesgoogle: take 'cloud' out of 'clouddebugger' for all of the languages here

google/devtools/clouddebugger/v2/clouddebugger_gapic.yaml
geigerj: That's what I have been doing. (copying the required fields into a new flattening section when manual edits bring the number of fields below the threshold)

google/devtools/clouddebugger/v2/clouddebugger_gapic.yaml
geigerj: AFAIK the current strategy is that the only time a non-required parameter goes in the flattening groups is it is a oneof; in that case, create one flattening group for each possible type of the oneof.

garrettjonesgoogle: *actually*, my strategy for Java so far has been to simply remove the flattening if there is a oneof parameter. I am concerned about the explosion of number of parameters. I think it might be better for callers to just construct a request and use the request-based method.

geigerj: Ah, OK. I think I got this impression from [RunQuery in datastore](https://github.com/googleapis/googleapis/blob/master/google/datastore/v1/datastore_gapic.yaml#L52), which it looks like @chrisdunelm configured. We should probably resolve this offline/separately from this PR.

google/devtools/clouddebugger/v2/clouddebugger_gapic.yaml
jcanizales: Done. Left C# because the spec doc says the package should be just the proto package, capitalized.

bjwatson:
@geigerj @garrettjonesgoogle Has the required/optional (and flattening) parameters been reviewed? I'm not going to have the time soon enough to review it myself, and I don't want to be a blocker on this PR. I'm currently requesting changes, and the only way I know to drop that lock is to approve this PR.

google/devtools/clouddebugger/v2/clouddebugger_gapic.yaml
geigerj: Please remove this param. (In this case, flattening == required)

google/devtools/clouddebugger/v2/clouddebugger_gapic.yaml
geigerj: Same thing -- make flattening == required fields

jcanizales:
I have one more change to make, to fix generation of flattening groups, and then it'll be ready for you guys to review the specific fields.

google/devtools/clouddebugger/v2/clouddebugger_gapic.yaml
jcanizales: Done.

google/devtools/clouddebugger/v2/clouddebugger_gapic.yaml
jcanizales: A flattening section was now generated.

bjwatson:
Just a question of process: Why are you re-generating rather than simply doing some hand-edits to refine this GAPIC config? Aren't there hand-edits that need to be applied anyway to the generated results?

bjwatson:
LGTM if @geigerj agrees

jcanizales:
Not one hand edit so far :) Everything's generated with just the string `devtools/clouddebugger/v2` as an input.

jcanizales:
Now to write / edit tests for my local changes in toolkit, and send them for review.

bjwatson:
@jcanizales Ok, cool. I guess there's nothing unusual here for idempotency.

PLMK if you need any help with codegen.

jcanizales:
Yeah the heuristic for idempotency worked fine in this case.

bjwatson:
@jcanizales Are you waiting for anything else to merge this?

jcanizales:
Ah, no; I was just used to the gRPC workflow in which the reviewer merges when he's happy. Now that GitHub supports real code reviews, the author merging works too :)

PR #146, opened 2016-09-20
Script to filter service configs for publishing

jcanizales:


garrettjonesgoogle:
Scripts should go into artman instead of googleapis.

geigerj:
There is actually a feature request for an internal tool to do this. It needs to be a little smarter than this script because in practice the service config is typically merged from several source files.

jcanizales:
This is what I'm using currently and I thought other people can use it to save themselves manual work. Should we:
* Keep it until we have the smarter tool.
* Iterate the smarter tool from it.
* Ignore it until we have the smarter tool and each person filters their own way.

jcanizales:
Offline conversation with Jacob:
* The smarter tool should be integrated with Blaze, and just be part of the normal `api_publish` flow, so this script can't get us closer to that goal.
* The lack of `config_version` came from not merging another yaml. It should totally merge, because e.g. auth scopes might be overriden. I think that's trivial to add to this script, so I'm going to try that.
* If we keep this somewhere shared for the team, it's of no big use that it's open-source, so we might as well keep it internally.

bjwatson:
@jcanizales Can you move this script elsewhere and close this PR? The script seems useful in the short-term, but not everything that we need per Jacob's comments.

jcanizales:
Yeah I already submitted it to my experimental directory in the internal repo, and added comments in the doc pointing to them.

bjwatson:
Ok, cool.

PR #147, opened 2016-09-20
Script to correct timeouts in generated GAPIC configs

jcanizales:
I'm using this as part of a script that runs the whole pipeline to generate a GAPIC config. It's called after the output file is extracted from the downloaded .tar.gz.

I'm not sure where to put the overall script, but as this is an independent piece of it I guess other people can use it meanwhile.

### DIFF
--- a/tools/correct_gapic_config_timeouts.py
+++ b/tools/correct_gapic_config_timeouts.py
@@ -1,0 +1,30 @@
+#! /usr/bin/python
+
+# TODO(jcanizales): What license header to use?
+
+# Requires:
+# pip install ruamel.yaml
+#
+# ruamel.yaml lets you read and write back Yaml files preserving the comments
+# and a nice layout.
+#
+# Example usage:
+# ./correct_gapic_config_timeouts.py wrong.yaml > correct.yaml
+
+import fileinput
+import ruamel.yaml
+import sys
+
+input_str = "".join(fileinput.input())
+
+# TODO(jcanizales): Abort with a nice error message if the file doesn't exist or
+# is empty.
+
+dom = ruamel.yaml.round_trip_load(input_str)
+
+for interface in dom['interfaces']:
+    for retry_def in interface['retry_params_def']:
+        retry_def['initial_rpc_timeout_millis'] = 60000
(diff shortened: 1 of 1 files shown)


geigerj:
Please fix this in the GAPIC config generator rather than using a script: https://github.com/googleapis/toolkit/blob/master/src/main/java/com/google/api/codegen/config/RetryGenerator.java#L71

garrettjonesgoogle:
This script shouldn't exist at all; if the default value is wrong, correct it in https://github.com/googleapis/toolkit/blob/master/src/main/java/com/google/api/codegen/config/RetryGenerator.java .

jcanizales:
Hm, the change to 20s was done on purpose here: https://github.com/googleapis/toolkit/commit/9652cb012458e299e2077ad67e5472cba0dca2b3#diff-9e97bbb6fd5002f21f2a66633fa62d93

So which should it be?

garrettjonesgoogle:
I've never been sure :-) Why did you want it to be set to 60 s?

jcanizales:
@bjwatson's doing it by hand. Brian, why not 20?

bjwatson:
I don't really know either? I thought we've been using all 60s across the board for other APIs.

Agreed that this should not be scripted. We should update the default instead.

jcanizales:
Hehe so if we don't have reasons to prefer one over the other, can we just leave it this way? 20s sounds as reasonable to me as 60s.

bjwatson:
@garrettjonesgoogle Who made the decision on 20 vs. 60s to begin with? They should be looped into this discussion.

garrettjonesgoogle:
@bjwatson I made the decision, somewhat arbitrarily.

bjwatson:
@garrettjonesgoogle Ok, then let's stop changing the timeout unless we find a reason to keep doing so.

bjwatson:
@jcanizales Looks like this PR can be closed. Can someone add an agenda item to discuss this next week, or start an email thread to discuss with the team? We should make a decision and apply it across the board to all APIs (except for any that need a custom timeout).

jcanizales:
I'll start a thread.

bjwatson:
Thanks @jcanizales

PR #149, opened 2016-09-21
Monitoring: Removing flattening until we figure out how to deal with oneof

garrettjonesgoogle:


michaelbausor:
LGTM, although another workaround would be to create 3 flattening groups, each with name and one of the oneof elements.

garrettjonesgoogle:
True, but I am concerned about that as a general strategy; if there are two oneof fields, it could create a combinatorial explosion. Just requiring the caller to use a request builder might be the better option. I don't want to decide right at the moment, though.

PR #152, opened 2016-09-21
Add resource_name_format.proto

michaelbausor:


garrettjonesgoogle:
LGTM, but I think we should have another person take a look too.

michaelbausor:
@ethanbao PTAL when you have a chance, and please ping me if you have any questions.

The extensions defined in resource_name_format.proto will be used by a protoc plugin to generate types for formatted resource name fields, and add getters and setters to the generated proto classes.

michaelbausor:
Closing now while considering GAPIC config option, will reopen if required

PR #155, opened 2016-09-27
Add LRO service config. Update Pub/Sub. Remove unneeded context.

bjwatson:


google/cloud/speech/cloud_speech.yaml
```diff
@@ -12,18 +12,8 @@ documentation:
 
 apis:
 - name: google.cloud.speech.v1beta1.Speech
-# TODO get LRO API call support added for Speech
-#- name: google.longrunning.Operations
+- name: google.longrunning.Operations
```
garrettjonesgoogle: Won't this duplicate the Operations Api in speech?

bjwatson: Are you talking about the `import "google/longrunning/operations.proto";` in `cloud_speech.proto`?

bjwatson: @garrettjonesgoogle Rather than doing this, can we handle it the same as IAM: adding the LRO methods directly to the Speech service, and using the reroute feature?

garrettjonesgoogle: Yes that is possible, but it's not really the intended design of Operations - it's meant to be the same interface when used in multiple APIs, in contrast with the way IAM works.

geigerj: To confirm, is the intended usage that a user must explicitly install/import a separate LRO client in order to query operations if they are using, for example, Cloud Speech?

geigerj: If so, that makes sense to me, and we should revert this change to `cloud_speech.yaml`.

garrettjonesgoogle: Yes, it is intended usage to import a separate LRO client. That said, in Java, the plan is to provide an instance of that LRO client through an accessor on the speech client.

bjwatson: Oh, shoot. I should have checked in on this thread. I just filed https://github.com/googleapis/googleapis/pull/158 showing how LRO could be used with Speech, the same way IAM is used with Pub/Sub.

bjwatson: So I don't understand what's different between IAM and LRO as far as this design goes. Are the IAM methods in Pub/Sub modified or given context in some manner via being a mixin? IOW, do they behave somehow differently than they would if they were rolled in a separate GAPIC package?

geigerj: AFAIK the main difference is that when you use a service as a mixin, you get to redefine the HTTP bindings. For GAPIC, this means that we can generate PathTemplate helpers for the IAM methods. This makes sense because IAM operates over resources. In contrast, LRO operates over objects so it doesn't need helper methods into the resource model of the API.

bjwatson: Ok, I'll rework https://github.com/googleapis/googleapis/pull/158

PR #158, opened 2016-09-27
Add GAPIC config for LRO

bjwatson:


bjwatson:
@garrettjonesgoogle I have re-purposed this PR to just add the GAPIC config for LRO, so we can build GAPIC wrappers for it to assist the google-cloud teams.

gapic/api/artman_lro.yaml
```diff
@@ -0,0 +1,25 @@
+common:
+  api_name: google-longrunning-operations
```
garrettjonesgoogle: I think I would prefer "google-longrunning". It matches the directory it's declared in.

bjwatson: Fair enough. That also shortens it without resorting to an arbitrary abbreviation (e.g. `lro`).

bjwatson:
PTAL

garrettjonesgoogle:
LGTM

PR #164, opened 2016-09-30
Update api_config_pattern, enable batch for PHP

michaelbausor:


gapic/batch/common.yaml
```diff
@@ -1,5 +1,5 @@
 common:
   batch_apis: '*'
   batch_languages: '*'
-  api_config_pattern: ${REPOROOT}/googleapis/gapic/api/artman_*.yaml
+  api_config_pattern: ${REPOROOT}/googleapis/gapic/api/artman_${APINAME}.yaml
```
garrettjonesgoogle: I think `APINAME` is potentially confusing, because in the artman config files, `api_name` contains the fully-qualified id for the service (e.g. `google-pubsub-v1`), whereas you want `pubsub` here. Maybe call it `API_SHORT_NAME`.

michaelbausor:
api_config_pattern update based on https://github.com/googleapis/artman/pull/105

michaelbausor:
Updated PTAL

garrettjonesgoogle:
LGTM

PR #165, opened 2016-09-30
Renaming longrunning output packages

garrettjonesgoogle:


garrettjonesgoogle:
Switching to gax repos instead (per internal thread).

garrettjonesgoogle:
Another update - removing 'cloud' from the package names.

garrettjonesgoogle:
PS I'm looking for a new "LGTM" because the original changes are all overwritten now.

bjwatson:
I'm okay with what I see, but let's wait until Monday for feedback from anyone else. For example, @jmuk is busy with other things today.

jmuk:
So the decision is longrunning.Operations as a part of gax of each languages?

I'm good with this config, but please note that it would require manual edits for NodeJS (I can do that on my side).

garrettjonesgoogle:
Well, I consider this PR to be the ratification of the proposal :-)

geigerj:
+1, but please let @bjwatson give final approval.

bjwatson:
@garrettjonesgoogle LGTM

PR #166, opened 2016-09-30
Removed _path from a path descriptor to avoid duplication

landrito:
For example problems arise [here](https://github.com/googleapis/api-client-staging/blob/master/generated/nodejs/google-cloud-node/packages/monitoring/src/v3/metric_service_api.js#L165)


garrettjonesgoogle:
lol nice

PR #168, opened 2016-10-06
Use consistent pattern for api_name fields

bjwatson:
This directly determines Python package names, which is visible to our
users.

### DIFF
--- a/gapic/api/artman_genomics.yaml
+++ b/gapic/api/artman_genomics.yaml
@@ -1,5 +1,5 @@
 common:
-  api_name: google-genomics-v1
+  api_name: google-cloud-genomics-v1
   proto_gen_pkg_deps:
     - google-common-protos
     - google-iam-v1
--- a/gapic/api/artman_longrunning.yaml
+++ b/gapic/api/artman_longrunning.yaml
@@ -1,5 +1,5 @@
 common:
-  api_name: google-longrunning
+  api_name: google-longrunning-v1
   proto_gen_pkg_deps:
     - google-common-protos
   import_proto_path:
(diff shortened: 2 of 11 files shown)


gapic/api/artman_genomics.yaml
```diff
@@ -1,5 +1,5 @@
 common:
-  api_name: google-genomics-v1
+  api_name: google-cloud-genomics-v1
```
garrettjonesgoogle: Is Genomics really cloud?

gapic/api/artman_longrunning.yaml
```diff
@@ -1,5 +1,5 @@
 common:
-  api_name: google-longrunning
+  api_name: google-longrunning-v1
```
garrettjonesgoogle: I'm quite hesitant to use v1 here, since "v1" technically isn't taken for it yet. What if someone names the next major version of it "v1"?

gapic/api/artman_genomics.yaml
bjwatson: https://cloud.google.com/genomics/

gapic/api/artman_longrunning.yaml
bjwatson: v1 appears in the REST interface: https://github.com/googleapis/googleapis/blob/master/google/longrunning/operations.proto#L43

bjwatson: I guess this change only matters if the Python gRPC code for LRO is shipped as a separate package, rather than as part of `googleapis-common-protos`. However, I'm reluctant to do that since common-protos is already post-1.x.

The GAPIC portion will be merged into `google-gax` (the `gax-python` package on PyPI), so it won't need its own package name.

garrettjonesgoogle: Ahh ok, if it appears in the http option, that's good enough for me.

garrettjonesgoogle:
LGTM

PR #172, opened 2016-10-14
Add domain_layer_location for Go

pongad:


pongad:
@garrettjonesgoogle Not sure if you'd be interested in this?

google/logging/v2/logging_gapic.yaml
```diff
@@ -6,6 +6,7 @@ language_settings:
     package_name: google.cloud.gapic.logging.v2
   go:
     package_name: cloud.google.com/go/logging/apiv2
+    domain_layer_location: cloud.google.com/go/logging
```
jba: OK, not a dealbreaker, but in an ideal world we'd point to two packages, logging and logging/logadmin.

Just file as a bug or TODO, don't worry about it now.

garrettjonesgoogle: @jba , what is logging/logadmin?

jba: We split the handwritten logging client into two packages: logging, for writing logs, and logging/logadmin, for reading logs and the other stuff.

PR #177, opened 2016-10-21
Setting required_constructor_params for LRO, other

garrettjonesgoogle:
Other:

- Turning off batch generation for APIs that are not ready yet.
- Fixing the java package of Google Cloud Functions

### DIFF
--- a/gapic/api/artman_clouddebugger.yaml
+++ b/gapic/api/artman_clouddebugger.yaml
@@ -11,7 +11,6 @@ common:
   gapic_api_yaml:
     - ${REPOROOT}/googleapis/google/devtools/clouddebugger/v2/clouddebugger_gapic.yaml
   output_dir: ${REPOROOT}/artman/output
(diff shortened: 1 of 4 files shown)


gapic/api/artman_clouddebugger.yaml
```diff
@@ -11,7 +11,7 @@ common:
   gapic_api_yaml:
     - ${REPOROOT}/googleapis/google/devtools/clouddebugger/v2/clouddebugger_gapic.yaml
   output_dir: ${REPOROOT}/artman/output
-  enable_batch_generation: True
+  enable_batch_generation: False
```
michaelbausor: Nit: did you mean to set this to false? Or can it be deleted?

michaelbausor: I think delete it - having it be false suggests a reason. But don't have strong feelings, was more just checking whether there *was* a reason that I wasn't seeing.

PR #180, opened 2016-10-25
Adding license config to GAPIC config files

garrettjonesgoogle:
Note: google/longrunning/longrunning_gapic.yaml is the odd one out (BSD license instead of Apache).

Coordinates with https://github.com/googleapis/toolkit/pull/676 .


garrettjonesgoogle:
It's actually better to merge this before toolkit.

PR #184, opened 2016-10-28
Add auth section to monitoring service config

geigerj:


jskeet:
Obsolete due to #183 I assume?

PR #191, opened 2016-11-09
Resource names pubsub yaml

michaelbausor:


google/pubsub/v1/pubsub_gapic.yaml
```diff
@@ -83,7 +83,7 @@ interfaces:
     retry_params_name: default
     field_name_patterns:
       name: subscription
-      topic: topic
+      topic: topic_oneof
```
garrettjonesgoogle: How does this work - what does it apply to?

google/pubsub/v1/pubsub_gapic.yaml
```diff
@@ -450,3 +472,50 @@ interfaces:
...
+    topic: topic
+- message_name: ListTopicsRequest
+  field_entity_map:
+    project: project
+- message_name: ListTopicSubscriptionsRequest
+  field_entity_map:
+    topic: topic
+- message_name: ListTopicSubscriptionsResponse
+  field_entity_map:
+      subscriptions: subscription
```
garrettjonesgoogle: The indentation is wrong here.

michaelbausor: Fixed

michaelbausor:
Currently the collection definition is available at the top level, so we could use it right now. My feeling was to wait until we also have the updates to config gen before moving existing collections to the top level, so that regenerating the config doesn't become more difficult that necessary.

garrettjonesgoogle:
ok. LGTM

PR #192, opened 2016-11-10
Add smoke test config

shinfan:
The following APIs will have the smoke test config:
- Language
- Speech
- Vision
- ErrorReporting
- Trace
- Logging
- Monitoring
- Pubsub

 TODO: Support enum init values in gapic config: https://github.com/googleapis/toolkit/issues/733

### DIFF
--- a/google/cloud/speech/v1beta1/cloud_speech_gapic.yaml
+++ b/google/cloud/speech/v1beta1/cloud_speech_gapic.yaml
@@ -22,10 +22,10 @@ interfaces:
   smoke_test:
     method: SyncRecognize
     init_fields:
-      - config.language_code="en-US"
-      - config.sample_rate=44100
-      - config.encoding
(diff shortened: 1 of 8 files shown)


google/cloud/speech/v1beta1/cloud_speech_gapic.yaml
```diff
@@ -19,6 +19,13 @@ license_header:
   license_file: license-header-apache-2.0.txt
 interfaces:
 - name: google.cloud.speech.v1beta1.Speech
+  smoke_test:
+    method: SyncRecognize
+    init_fields:
+      - config.language_code="en-US"
```
garrettjonesgoogle: the lines starting with '-' should have the same number of spaces as the key line (so two spaces in from where you have it)

shinfan: Done

garrettjonesgoogle:
LGTM

PR #193, opened 2016-11-11
Create a simple overview of Service Config.

guptasu:
Endpoints and other products can link to it and build their own documentation referencing this readme.

### DIFF
--- a/google/api/README.md
+++ b/google/api/README.md
@@ -116,7 +116,7 @@ example, `versioning-*` suppresses all warnings related to the
 `versioning` group.
 
 Suppression directives attached to parent elements apply to all
-childs. For example, the above suppression when attached to the
+children. For example, the above suppression when attached to the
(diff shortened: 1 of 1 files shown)


google/api/README.md
```diff
@@ -1,5 +1,143 @@
...
+  }
+}
+```
+
+Instead of the rule name a wildcard (`*`) can be provided. For
+example, `versioning-*` suppresses all warnings related to the
+`versioning` group.
+
+Suppression directives attached to parent elements apply to all
+childs. For example, the above suppression when attached to the
```
garrettjonesgoogle: s/childs/children/

wora:
Do we plan to finish this one?

guptasu:
The documentation is now available under endpoints cloud documentation. https://cloud.google.com/endpoints/docs/grpc-service-config

PR #199, opened 2016-11-21
update speech config with lro

pongad:


garrettjonesgoogle:
Nice, I have this exact change locally, I just haven't pushed it yet :-P LGTM

pongad:
@garrettjonesgoogle Dennis Ritchie and Ken Thompson worked together so well that the one time they got their wires crossed, they end up writing the same function byte-for-byte identical. I guess we can baby-step with a config for now.

PR #201, opened 2016-11-22
Removing collection config for operation_path

garrettjonesgoogle:
The APIs that use longrunning as a mix-in can use any path
of their choosing.


michaelbausor:
LGTM

PR #203, opened 2016-11-22
Removing validation for log, parent, and sync resource names

garrettjonesgoogle:
* metric_name still has only one format, so it will still be validated.

### DIFF
--- a/google/logging/v2/logging_gapic.yaml
+++ b/google/logging/v2/logging_gapic.yaml
@@ -62,6 +62,7 @@ interfaces:
     timeout_millis: 30000
     field_name_patterns:
       parent: parent
+    resource_name_treatment: NONE
   - name: GetSink
     flattening:
       groups:
@@ -75,6 +76,7 @@ interfaces:
     timeout_millis: 30000
     field_name_patterns:
       sink_name: sink
+    resource_name_treatment: NONE
   - name: CreateSink
     flattening:
       groups:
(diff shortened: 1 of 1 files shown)


google/logging/v2/logging_gapic.yaml
```diff
@@ -161,11 +166,14 @@ interfaces:
     timeout_millis: 30000
     field_name_patterns:
       parent: parent
+    resource_name_treatment: NONE
   - name: GetLogMetric
     flattening:
       groups:
       - parameters:
         - metric_name
+        parameter_resource_name_treatment:
```
michaelbausor: Why set the method-level setting to NONE and override the only flattened parameter? Is this necessary for the unit tests?

garrettjonesgoogle: It's easier to be consistent this way - I think future maintainers are less likely to make a mistake if all method-level settings are NONE.

michaelbausor: Fair enough

michaelbausor:
LGTM

swcloud:
@garrettjonesgoogle @michaelbausor Is code regeneration needed by this change?

Maiconcoringa (approved):

Maiconcoringa (approved):

Maiconcoringa (approved):

PR #204, opened 2016-11-23
Add configuration for PackageMetadataGen

geigerj:
Copies some files from Packman to googleapis, and passes them to
Artman's pipeline kwargs.

Once all languages stop using Packman, we can delete the duplicate copies in that repo.


garrettjonesgoogle:
LGTM

PR #209, opened 2016-12-01
Add "cloud" to IAM API name

geigerj:
If IAM is considered a "Cloud" API, we should make this naming change.

@bjwatson If it isn't, we need to adjust the Python package-renaming logic not to apply to IAM protos and gRPC generated code.


garrettjonesgoogle:
This one isn't a cloud API I don't think - non-cloud services could use IAM as a mixin.

geigerj:
According to @bjwatson this should actually be folded into common-protos.

PR #215, opened 2016-12-09
Adopt new artman naming configuration

geigerj:
Corresponds to https://github.com/googleapis/artman/pull/123

### DIFF
--- a/gapic/api/artman_bigtable.yaml
+++ b/gapic/api/artman_bigtable.yaml
@@ -1,6 +1,6 @@
 common:
-  short_name: bigtable
-  version: v2
+  api_name: bigtable
+  api_version: v2
   proto_gen_pkg_deps:
     - google-common-protos
   import_proto_path:
--- a/gapic/api/artman_bigtable.yaml
+++ b/gapic/api/artman_bigtable.yaml
@@ -1,6 +1,7 @@
 common:
   api_name: bigtable
   api_version: v2
+  organization_name: google-cloud
   proto_gen_pkg_deps:
     - google-common-protos
   import_proto_path:
--- a/gapic/api/artman_bigtable.yaml
+++ b/gapic/api/artman_bigtable.yaml
@@ -1,5 +1,6 @@
 common:
-  api_name: google-cloud-bigtable-v2
+  short_name: bigtable
+  version: v2
   proto_gen_pkg_deps:
     - google-common-protos
   import_proto_path:
--- a/gapic/api/artman_longrunning.yaml
+++ b/gapic/api/artman_longrunning.yaml
@@ -1,5 +1,7 @@
 common:
-  api_name: google-longrunning-v1
+  short_name: longrunning
+  version: v1
+  is_cloud_api: False
   proto_gen_pkg_deps:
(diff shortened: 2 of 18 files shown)


gapic/api/artman_bigtable.yaml
```diff
@@ -1,5 +1,6 @@
 common:
-  api_name: google-cloud-bigtable-v2
+  short_name: bigtable
+  version: v2
```
garrettjonesgoogle: potentially 'api_short_name' and 'api_version' would be clearer.

gapic/api/artman_longrunning.yaml
```diff
@@ -1,5 +1,7 @@
 common:
-  api_name: google-longrunning-v1
+  short_name: longrunning
+  version: v1
+  is_cloud_api: False
```
garrettjonesgoogle: related to my comment on artman - I think it would be better to have `org_id: google_cloud`

geigerj:
@garrettjonesgoogle I'll update this PR once googleapis/artman#123 so that the discussion can happen in one place.

gapic/api/artman_bigtable.yaml
geigerj: `version` -> `api_version` makes sense to me because we need to distinguish API/client lib version. Done on the artman PR.

`api_short_name` seems verbose given that this is actually now the only API name in the artman config; how about just calling this `api_name` now? (To be distinguished from [the API `title` in the service config](https://github.com/googleapis/googleapis/blob/master/google/logging/logging.yaml#L6).)

garrettjonesgoogle: Yeah I think you're right that it's verbose - `api_name` makes sense. Potentially the full name in artman should be `api_full_name` (not relevant to this config, but relevant to the overall naming).

geigerj: Done in the artman PR.

geigerj:
@garrettjonesgoogle Updated this PR as well, PTAL?

gapic/lang/common.yaml
```diff
@@ -5,6 +5,7 @@ common:
   package_dependencies_yaml: ${REPOROOT}/googleapis/gapic/packaging/dependencies.yaml
   package_defaults_yaml: ${REPOROOT}/googleapis/gapic/packaging/api_defaults.yaml
   common_protos_yaml: ${REPOROOT}/googleapis/gapic/packaging/common_protos.yaml
+  organization_name: google-cloud
```
garrettjonesgoogle: I think I would prefer to put `google-cloud` in all of the artman yaml files - it's something we'll have to add everywhere later anyway as we expand to other organizations, which will overwhelm google-cloud in the number of APIs.

geigerj: Actually, I realize this doesn't work; according to the current recommended artman `USAGE`, the values in common.yaml override the values in artman_{api}.yaml.

geigerj: Done

garrettjonesgoogle:
LGTM

PR #217, opened 2016-12-09
Update gapic yaml with streaming pull settings

michaelbausor:


google/pubsub/v1/pubsub_gapic.yaml
```diff
@@ -186,6 +186,20 @@ interfaces:
...
       subscription: subscription
     timeout_millis: 60000
+  - name: StreamingPull
+    required_fields:
+    - subscription
+    - stream_ack_deadline_seconds
+    grpc_streaming:
+      response:
+        resources_field: received_messages
+    request_object_method: true
```
garrettjonesgoogle: a request object method doesn't make sense for a streaming method. Maybe we should even make this an error.

michaelbausor: Fixed, and created issue: https://github.com/googleapis/toolkit/issues/850

michaelbausor:
PTAL

garrettjonesgoogle:
LGTM

PR #218, opened 2016-12-12
Update logging config with ListLogs method config

michaelbausor:


google/logging/v2/logging_gapic.yaml
```diff
@@ -380,6 +380,27 @@ interfaces:
...
+    page_streaming:
+      request:
+        page_size_field: page_size
+        token_field: page_token
+      response:
+        token_field: next_page_token
+        resources_field: log_names
+    retry_codes_name: idempotent
+    retry_params_name: default
+    field_name_patterns:
```
jskeet: Hmm... the URL pattern in the method option suggests it could just be a project, but the docs suggest it could be an org. This looks like the safe bet.

garrettjonesgoogle: The proto in googleapis doesn't reflect the full list in the internal yaml file. (That is a separate issue that should be fixed.)

PR #221, opened 2016-12-13
Add bigtable admin API

michaelbausor:
Add the service, artman and gapic yaml files for the bigtable admin API

### DIFF
--- a/google/bigtable/admin/v2/bigtable_admin_gapic.yaml
+++ b/google/bigtable/admin/v2/bigtable_admin_gapic.yaml
@@ -1,0 +1,300 @@
+type: com.google.api.codegen.ConfigProto
+language_settings:
+  java:
+    package_name: com.google.cloud.bigtable.admin.spi.v2
+  python:
+    package_name: google.cloud.gapic.bigtable.admin.v2
+  go:
+    package_name: cloud.google.com/go/bigtable/admin/apiv2
+  csharp:
+    package_name: Google.Bigtable.Admin.V2
+  ruby:
+    package_name: Google::Cloud::Bigtable::Admin::V2
+  php:
+    package_name: Google\Cloud\Bigtable\Admin\V2
+  nodejs:
+    package_name: admin.v2
+license_header:
+  copyright_file: copyright-google.txt
+  license_file: license-header-apache-2.0.txt
+interfaces:
+- name: google.bigtable.admin.v2.BigtableInstanceAdmin
+  collections:
+  - name_pattern: projects/{project}
+    entity_name: project
+  - name_pattern: projects/{project}/instances/{instance}
+    entity_name: instance
+  - name_pattern: projects/{project}/instances/{instance}/clusters/{cluster}
+    entity_name: cluster
+  retry_codes_def:
+  - name: idempotent
+    retry_codes:
+    - UNAVAILABLE
+    - DEADLINE_EXCEEDED
+  - name: non_idempotent
+    retry_codes: []
+  retry_params_def:
+  - name: default
+    initial_retry_delay_millis: 100
+    retry_delay_multiplier: 1.3
+    max_retry_delay_millis: 60000
+    initial_rpc_timeout_millis: 20000
+    rpc_timeout_multiplier: 1
+    max_rpc_timeout_millis: 20000
+    total_timeout_millis: 600000
+  methods:
+  - name: CreateInstance
+    flattening:
+      groups:
+      - parameters:
+        - parent
+        - instance_id
+        - instance
+        - clusters
+    required_fields:
+    - parent
+    - instance_id
+    - instance
+    - clusters
+    request_object_method: true
+    retry_codes_name: non_idempotent
+    retry_params_name: default
+    field_name_patterns:
+      parent: project
+    timeout_millis: 60000
+  - name: GetInstance
+    flattening:
+      groups:
+      - parameters:
+        - name
+    required_fields:
+    - name
+    request_object_method: false
+    retry_codes_name: idempotent
+    retry_params_name: default
+    field_name_patterns:
+      name: instance
+    timeout_millis: 60000
+  - name: ListInstances
+    flattening:
+      groups:
+      - parameters:
+        - parent
+    required_fields:
+    - parent
+    request_object_method: true
+    retry_codes_name: idempotent
+    retry_params_name: default
+    field_name_patterns:
+      parent: project
+    timeout_millis: 60000
+  - name: UpdateInstance
+    flattening:
+      groups:
+      - parameters:
+        - name
+        - display_name
+    required_fields:
+    - name
+    - display_name
+    request_object_method: true
+    retry_codes_name: idempotent
+    retry_params_name: default
+    field_name_patterns:
+      name: instance
+    timeout_millis: 60000
+  - name: DeleteInstance
+    flattening:
+      groups:
+      - parameters:
+        - name
+    required_fields:
+    - name
+    request_object_method: false
+    retry_codes_name: idempotent
+    retry_params_name: default
+    field_name_patterns:
+      name: instance
+    timeout_millis: 60000
+  - name: CreateCluster
+    flattening:
+      groups:
+      - parameters:
+        - parent
+        - cluster_id
+        - cluster
+    required_fields:
+    - parent
+    - cluster_id
+    - cluster
+    request_object_method: true
+    retry_codes_name: non_idempotent
+    retry_params_name: default
+    field_name_patterns:
+      parent: instance
+    timeout_millis: 60000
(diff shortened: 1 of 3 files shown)


garrettjonesgoogle:
Where is the proto that this corresponds to?

michaelbausor:
The instance admin and table admin protos here: https://github.com/googleapis/googleapis/tree/master/google/bigtable/admin/v2

google/bigtable/admin/v2/bigtable_admin_gapic.yaml
```diff
@@ -0,0 +1,300 @@
...
+        - cluster
+    required_fields:
+    - parent
+    - cluster_id
+    - cluster
+    request_object_method: true
+    retry_codes_name: non_idempotent
+    retry_params_name: default
+    field_name_patterns:
+      parent: instance
```
garrettjonesgoogle: I think this should be `cluster` instead of `instance`

google/bigtable/admin/v2/bigtable_admin_gapic.yaml
```diff
@@ -0,0 +1,300 @@
...
+        - name
+        - serve_nodes
+    required_fields:
+    - name
+    - serve_nodes
+    request_object_method: true
+    retry_codes_name: idempotent
+    retry_params_name: default
+    field_name_patterns:
+      name: cluster
```
garrettjonesgoogle: Did you consider doing a resource name for `location`? If we do, we'd probably want to ask the BigTable team how flexible it should be.

google/bigtable/admin/v2/bigtable_admin_gapic.yaml
```diff
@@ -0,0 +1,300 @@
...
+  - name: default
+    initial_retry_delay_millis: 100
+    retry_delay_multiplier: 1.3
+    max_retry_delay_millis: 60000
+    initial_rpc_timeout_millis: 20000
+    rpc_timeout_multiplier: 1
+    max_rpc_timeout_millis: 20000
+    total_timeout_millis: 600000
+  methods:
+  - name: CreateInstance
```
garrettjonesgoogle: This method needs LRO config.

google/bigtable/admin/v2/bigtable_admin_gapic.yaml
```diff
@@ -0,0 +1,300 @@
...
+        - name
+    required_fields:
+    - name
+    request_object_method: false
+    retry_codes_name: idempotent
+    retry_params_name: default
+    field_name_patterns:
+      name: instance
+    timeout_millis: 60000
+  - name: CreateCluster
```
garrettjonesgoogle: This method needs LRO config.

google/bigtable/admin/v2/bigtable_admin_gapic.yaml
```diff
@@ -0,0 +1,300 @@
...
+        - parent
+    required_fields:
+    - parent
+    request_object_method: true
+    retry_codes_name: idempotent
+    retry_params_name: default
+    field_name_patterns:
+      parent: instance
+    timeout_millis: 60000
+  - name: UpdateCluster
```
garrettjonesgoogle: This method needs LRO config.

google/bigtable/admin/v2/bigtable_admin_gapic.yaml
```diff
@@ -0,0 +1,300 @@
...
+        - name
+    required_fields:
+    - name
+    request_object_method: false
+    retry_codes_name: idempotent
+    retry_params_name: default
+    field_name_patterns:
+      name: instance
+    timeout_millis: 60000
+  - name: ListInstances
```
garrettjonesgoogle: Do we have to do anything special due to the list methods here not containing a page_size field? This method is also missing page streaming configuration. (I'm guessing that's because of the missing page_size)

google/bigtable/admin/v2/bigtable_admin_gapic.yaml
```diff
@@ -0,0 +1,300 @@
...
+    timeout_millis: 60000
+  - name: UpdateInstance
+    flattening:
+      groups:
+      - parameters:
+        - name
+        - display_name
+    required_fields:
+    - name
+    - display_name
```
garrettjonesgoogle: I noticed that the `state` field on `Cluster` is `@OutputOnly`, but not on Instance. Should we have the BigTable team clarify?

google/bigtable/admin/v2/bigtable_admin_gapic.yaml
```diff
@@ -0,0 +1,300 @@
...
+        - instance_id
+        - instance
+        - clusters
+    required_fields:
+    - parent
+    - instance_id
+    - instance
+    - clusters
+    request_object_method: true
+    retry_codes_name: non_idempotent
```
garrettjonesgoogle: I suspect this is idempotent because the unique identifier for the object is provided by the user.

google/bigtable/admin/v2/bigtable_admin_gapic.yaml
```diff
@@ -0,0 +1,300 @@
...
+      - parameters:
+        - parent
+        - cluster_id
+        - cluster
+    required_fields:
+    - parent
+    - cluster_id
+    - cluster
+    request_object_method: true
+    retry_codes_name: non_idempotent
```
garrettjonesgoogle: I suspect this is idempotent because the unique identifier for the object is provided by the user.

google/bigtable/admin/v2/bigtable_admin_gapic.yaml
```diff
@@ -0,0 +1,300 @@
...
+      - parameters:
+        - parent
+        - table_id
+        - table
+        - initial_splits
+    required_fields:
+    - parent
+    - table_id
+    - table
+    - initial_splits
```
garrettjonesgoogle: `initial_splits` is optional

google/bigtable/admin/v2/bigtable_admin_gapic.yaml
```diff
@@ -0,0 +1,300 @@
...
+  - name: default
+    initial_retry_delay_millis: 100
+    retry_delay_multiplier: 1.3
+    max_retry_delay_millis: 60000
+    initial_rpc_timeout_millis: 20000
+    rpc_timeout_multiplier: 1
+    max_rpc_timeout_millis: 20000
+    total_timeout_millis: 600000
+  methods:
+  - name: CreateTable
```
garrettjonesgoogle: I suspect this is idempotent because the unique identifier for the object is provided by the user.

google/bigtable/admin/v2/bigtable_admin_gapic.yaml
michaelbausor: The page streaming config is missing because there are two repeated fields in the response object. I looked at adding the page streaming config, but I am not sure we want to, as it makes accessing the other field more difficult and much less obvious. What do you think?

google/bigtable/admin/v2/bigtable_admin_gapic.yaml
```diff
@@ -0,0 +1,300 @@
...
+  ruby:
+    package_name: Google::Cloud::Bigtable::Admin::V2
+  php:
+    package_name: Google\Cloud\Bigtable\Admin\V2
+  nodejs:
+    package_name: admin.v2
+license_header:
+  copyright_file: copyright-google.txt
+  license_file: license-header-apache-2.0.txt
+interfaces:
```
michaelbausor: I didn't add resource names because the main bigtable doesn't have them yet - I thought it would make sense to add them together.

google/bigtable/admin/v2/bigtable_admin_gapic.yaml
michaelbausor: Yes we should get clarification, although it seems unlikely that the state can be set by the client in a meaningful way

google/bigtable/admin/v2/bigtable_admin_gapic.yaml
michaelbausor: I'm not sure, the http binding on the rpc suggests instance, and the cluster id is also specified separately in the request, so I think this should be instance and the comment on the field is wrong. But I will clarify with bigtable.

google/bigtable/admin/v2/bigtable_admin_gapic.yaml
garrettjonesgoogle: I didn't notice the two repeated fields, good point. I'm not sure about this one, I'll have to think about it.

lukesneeringer:
I am guessing that this is dead. Please feel free to re-open if I am wrong.

Issue #187, opened 2016-11-01
Add protos and generate gRPC code for Audit Log Datatypes


bjwatson:
###What:

Add the proto files for the audit types discussed here: https://cloud.google.com/logging/docs/audit/api/

Generate gRPC code for the appropriate languages:

* [] Python

###Why:

This allows audit log data to be passed as a `proto_payload` field, as documented here: https://cloud.google.com/logging/docs/view/logs_index#audit-logs

bjwatson:
@jba I think you said that this has already been done for Go, correct?

bjwatson:
The AuditLog proto has already been published: https://github.com/googleapis/googleapis/blob/master/google/cloud/audit/audit_log.proto

We just need to make sure it's up to date, since it was last published two months ago.

jba:
Yes, I linked the generated proto
<https://github.com/GoogleCloudPlatform/google-cloud-go/blob/master/logging/logadmin/logadmin.go#L49>
into the Go package that reads logs.

bjwatson:
Awesome, I checked the box for Go in this issue.

jba:
What about the four additional types listed at the bottom of that doc:
```
google.appengine.v1beta4.AuditData
google.appengine.legacy.AuditData
google.cloud.bigquery.logging.v1.AuditData
google.iam.v1.logging.AuditData
```
We should probably include those too, right?

bjwatson:
Yep, good point. We need to publish all of those to https://github.com/googleapis/googleapis

bjwatson:
I updated the comment, and unchecked Go since it doesn't yet have those remaining types.

bjwatson:
Assigning @geigerj for Python part. Other language leads please address your own languages.

FYI, we are planning to release one new package containing both this and the [GAE RequestLog object](https://github.com/GoogleCloudPlatform/google-cloud-python/issues/2572). Please let us know if you think they should be separate packages.

bjwatson:
@nkubala we'll probably fix this for Python at the same time as https://github.com/GoogleCloudPlatform/google-cloud-python/issues/2572.

geigerj:
@bjwatson: @shinfan and I are still working on getting these files open sourced. I'm going to go ahead separate this out from GoogleCloudPlatform/google-cloud-python#2572 in order not to block the other issue. We can add the AuditData types to the package created for that issue when they're open sourced.

bjwatson:
[...]

PR #226, opened 2016-12-21
Add configuration for Logging external types

geigerj:
Note that because Python gRPC metadata generation now uses toolkit and
is build on api-compiler, a service config file is required. This
config is used only to generate the gRPC package metadata.

Updates https://github.com/GoogleCloudPlatform/google-cloud-python/issues/2572
Updates https://github.com/googleapis/googleapis/issues/187

cc: @ethanbao regarding addition of this new service config

### DIFF
--- a/gapic/core/artman_logging_types.yaml
+++ b/gapic/core/artman_logging_types.yaml
@@ -1,0 +1,16 @@
+common:
+  api_name: service-logging-types
+  api_version: v1
+  organization_name: google
+  proto_gen_pkg_deps:
+    - google-common-protos
+  import_proto_path:
+    - ${REPOROOT}/googleapis
+  src_proto_path:
+    # Remaining AuditData types should be added here. See
+    #   https://github.com/googleapis/googleapis/issues/187
+    - ${REPOROOT}/googleapis/google/appengine/logging/v1
+    - ${REPOROOT}/googleapis/google/appengine/legacy
+  service_yaml:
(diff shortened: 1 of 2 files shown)


gapic/core/artman_logging_types.yaml
```diff
@@ -0,0 +1,16 @@
+common:
+  api_name: service-logging-types
```
bjwatson: `logging-external-types` or `logging-extra-types` seems like a better name for this package. I stumble on the meaning of the `service` token when mentally parsing `grpc-google-service-logging-types-v1`. Also, it would be good if this is consistent with the service config filename.

gapic/core/artman_logging_types.yaml
```diff
@@ -0,0 +1,16 @@
...
+  api_version: v1
+  organization_name: google
+  proto_gen_pkg_deps:
+    - google-common-protos
+  import_proto_path:
+    - ${REPOROOT}/googleapis
+  src_proto_path:
+    # Remaining AuditData types should be added here. See
```
bjwatson: Make this a TODO comment

bjwatson: You could also spec out what the other lines will be and just comment them, so it's like a checklist for what remains to remove this TODO comment.

lukesneeringer:
I am guessing that this is dead. Please feel free to re-open if I am wrong.

PR #230, opened 2017-01-05
Add Python GAX package dependency

geigerj:
Corresponds to https://github.com/googleapis/artman/pull/130


geigerj:
@shinfan This small change is required before I submit  googleapis/artman#130, could you take a look when you have a chance?

shinfan:
LGTM

PR #231, opened 2017-01-06
Increase the total timeout for streamingRecognize to 190 seconds.

jmuk:
There are some error reports about gRPC deadline is too short
(see https://github.com/GoogleCloudPlatform/google-cloud-node/issues/1894)

According to https://cloud.google.com/speech/limits#content,
the streamingRecognize method has the limit of ~1 minutes at API level,
and based on the error message, the total timeout should be around
185 seconds (60 seconds * 3 + 5 seconds). To make a little buffer,
this PR sets the timeout to be 190 seconds.

### DIFF
--- a/google/cloud/speech/v1beta1/cloud_speech_gapic.yaml
+++ b/google/cloud/speech/v1beta1/cloud_speech_gapic.yaml
@@ -77,4 +77,4 @@ interfaces:
   - name: StreamingRecognize
     retry_codes_name: non_idempotent
     retry_params_name: default
-    timeout_millis: 60000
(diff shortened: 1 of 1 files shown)


jmuk:
> SyncRecognize has the same 1-minute limit. Should you change that as well?

As far as I tested locally, it successfully finished recognition for ~55 seconds audio with the current timeout.

> Even if increasing the time-out here, it won't change the audio length (1 minute) API limit.

Right. That should be documented on the documentation on each google-cloud-[lang] repositories.

garrettjonesgoogle:
LGTM

Issue #222, opened 2016-12-15
LANGUAGE variable in Makefile elided by LANGUAGE environment variable


allquixotic:
The LANGUAGE variable in the Makefile is elided by the LANGUAGE environment variable. 

On many systems (including Ubuntu and MSYS2), the LANGUAGE environment variable is defined, e.g. as "en_US" or something like that.

This leads to `make` producing output like this:

```
$ make
mkdir -p ./gens
protoc --proto_path=.:/usr/local/include --en_US_out=./gens --grpc_out=./gens --plugin=protoc-gen-grpc=/usr/local/bin/grpc_en_US_plugin google/api/annotations.proto
protoc-gen-en_US: program not found or is not executable
--en_US_out: protoc-gen-en_US: Plugin failed with status code 1.
Makefile:41: recipe for target 'google/api/annotations.pb.cc' failed
make: *** [google/api/annotations.pb.cc] Error 1
```

Changing the variable name in the Makefile to something different - anything not already defined in the environment - fixes the issue.

bloodcarter:
+1

wora:
According to POSIX standard, http://pubs.opengroup.org/onlinepubs/
009695399/basedefs/xbd_chap08.html, LANG is the standard environment
variable.

This  Makefile is intended to verify the entire repo is buildable. It is
not intended to be used as "make" without a language.

allquixotic:
From the README.md:

 > To generate gRPC source code for Google APIs in this repository, you first need to install both Protocol Buffers and gRPC on your local machine. Then you can run `make all` in this directory to generate the code.

Running `make all` is synonymous with running `make`, so the intention of the Makefile is not at all clear from the readme.

Nowhere does it say that the user is responsible for specifying a language, though in practice that's what ends up happening (or the user has to edit the Makefile, which is even less friendly).

Suggest updating the docs to provide the recommended way for users to generate the bindings for e.g. C++ and C#.

geigerj:
If the makefile was originally just intended to ensure that the entire repo is buildable, as @wora said, we no longer need it; there are other processes currently used to check this, and no one currently maintains that file. We will remove the makefile. In the meantime, please invoke `protoc` directly.

I will update the docs and remove this file.

wora:
I will update the doc.

wora:
Please keep the Makefile and I will fix it. We need to demonstrate the repo
can be easily built with absolutely minimum dependency. If it is hard to
[...]

PR #232, opened 2017-01-08
Clarify the purpose of Makefile

wora:
I also improve the documentation a bit.


wora:
This will address #222.

PR #234, opened 2017-01-09
Allow specifying protoc binary with environment variable.

kskalski:


wora:
LGTM

We intend to let people customize the Makefile behavior. Personally, I think the variable should be passed via command line, not environment variable.

PR #235, opened 2017-01-10
Update config files to use consistent terminology

geigerj:
This is a prerequisite to https://github.com/googleapis/artman/pull/131

- Also removes some old `packman` fields that are not currently used by the `toolkit` metadata generation.
- Also copies some previously hardcoded PHP configuration that was deleted from `toolkit` in https://github.com/googleapis/toolkit/pull/870 (FYI @michaelbausor)

### DIFF
--- a/gapic/packaging/api_defaults.yaml
+++ b/gapic/packaging/api_defaults.yaml
@@ -19,7 +19,7 @@ author: Google Inc
 email: googleapis-packages@google.com
 homepage: https://github.com/googleapis/googleapis
 license: Apache-2.0
-package_version:
+generated_package_version:
   # TODO: Python must go to 1.0.20 or 1.1.0 when we GA, due to mistakenly
   #       publishing some 1.0.x versions to pypi already. They are now
   #       deleted but files still exist on the pypi servers and can prevent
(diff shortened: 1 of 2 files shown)


gapic/packaging/api_defaults.yaml
```diff
@@ -16,14 +16,18 @@
 # Provides default values used as metadata when constructing various packages.
 
 author: Google Inc
-description: a grpc-based api
 email: googleapis-packages@google.com
-github_user_uri: https://github.com/googleapis
 homepage: https://github.com/googleapis/googleapis
 license: Apache-2.0
-semver:
+package_version:
```
garrettjonesgoogle: maybe `api_package_version`?

geigerj: Hm, to me that's more confusing, because `api_version` refers to something unrelated to this configuration value.

garrettjonesgoogle: Hmm, maybe `this_package_version` or `generated_package_version` then? I think just `package_version` is a context-less description and it seems like it could apply to anything.

geigerj: Changed to `generated_package_version`

garrettjonesgoogle:
LGTM

PR #239, opened 2017-01-13
Change grpc- prefix to proto-.

lukesneeringer:
This PR bumps version numbers.

It must be co-merged alongside:
  * googleapis/artman#132
  * googleapis/toolkit#930


geigerj:
I think we can close this due to #240 

FYI: @eoogbe let's hold off actually pushing new packages to PyPI until @lukesneeringer's changes are merged to avoid having to immediately republish the packages.

PR #252, opened 2017-02-10
Java: pull -> public, publish -> deprecated

garrettjonesgoogle:


shinfan:
LGTM. Just one quick question, since publish is marked as deprecated, what is the new method? (It is not clear to me which new method should be used from the doc)

garrettjonesgoogle:
`Publisher` should be used instead. We have a note in the class doc, but not yet on the method doc (we didn't realize this had to be visible until the last minute - the deprecated handwritten classes still need to access it).

PR #254, opened 2017-02-13
move go artifacts to api-client-staging

pongad:
Many languages put the generated artifacts in the google-cloud-{lang}
directory under REPOROOT.
One unusual thing about Go is that the toolchain assumes all Go code to
live inside a directory named by env var GOPATH.
This path is different on each machine,
so we don't have a good default place to put the artifacts.

Knowing this, we might as well make it easier to put things into
staging repo.


garrettjonesgoogle:
Staging is supposed to be a separate target location from the final repo dir. Setting them to the same thing might cause funny things to happen if anything assumes the staging dir and final repo dirs are different.

PR #257, opened 2017-02-14
Updates to Spanner service configs

bjwatson:


bjwatson:
@garrettjonesgoogle @jonparrott Can you take a look at the auth scope changes for the admin APIs? As I was going through the latest service config, it looked like these are what we actually need here, but I want to make sure. Also, what would be the impact on the codegen? Anything, since we just require all scopes anyway?

garrettjonesgoogle:
I'm not aware of us using the value under `selector`.

bjwatson:
Ok, so this is really just a no-op change (albeit one that seems more factually correct).

@lukesneeringer You have the LGTM from @tomayles.

bjwatson:
And by @lukesneeringer, I actually meant myself. Too many balls in the air right now. :)

PR #263, opened 2017-02-23
Updates to pubsub gapic yaml.

landrito:


google/pubsub/v1/pubsub_gapic.yaml
```diff
@@ -221,6 +238,71 @@ interfaces:
...
+    request_object_method: false
+    retry_codes_name: idempotent
+    retry_params_name: default
+    field_name_patterns:
+      snapshot: snapshot
+    timeout_millis: 60000
+  - name: Seek
+    flattening:
+      groups:
+      - parameters:
```
landrito: I'd like some extra attention to this part. I'm not 100% on how we format this for oneofs.

jskeet: I don't think it's entirely sorted for statically typed languages. We considered it ages ago, but didn't finish the work. I've raised https://github.com/googleapis/toolkit/issues/1057 to cover this.

google/pubsub/v1/pubsub_gapic.yaml
```diff
@@ -77,6 +79,8 @@ interfaces:
         - topic
         - push_config
         - ack_deadline_seconds
+        - retain_acked_messages
+        - message_retention_duration
```
garrettjonesgoogle: It is a breaking change to add parameters to a flattening group.

google/pubsub/v1/pubsub_gapic.yaml
```diff
@@ -102,6 +106,21 @@ interfaces:
...
+        - subscription
+        - update_mask
+    required_fields:
+    - subscription
+    - update_mask
+    request_object_method: true
+    retry_codes_name: non_idempotent
+    retry_params_name: default
+    field_name_patterns:
+      subscription.name: subscription
```
garrettjonesgoogle: I don't think you can do subfields for field_name_patterns.

google/pubsub/v1/pubsub_gapic.yaml
```diff
@@ -221,6 +240,71 @@ interfaces:
...
+    timeout_millis: 60000
+  - name: Seek
+    flattening:
+      groups:
+      - parameters:
+        - subscription
+        - time
+      - parameters:
+        - subscription
+        - snapshot
```
garrettjonesgoogle: What's the reason for two flattening groups?

google/pubsub/v1/pubsub_gapic.yaml
landrito: Would you suggest creating a new flattening group or just using the previous?

google/pubsub/v1/pubsub_gapic.yaml
landrito: This is because of the oneof containing time and snapshot. Admittedly this is what I am not 100% sure of as said in my previous comment.

I am assuming that with two flattening groups, two methods will be generated.

`Seek(String subscription, Time time)` and `Seek(String subscription, String snapshot)`

google/pubsub/v1/pubsub_gapic.yaml
landrito: This was from the config generator so I assume it is correct.

google/pubsub/v1/pubsub_gapic.yaml
garrettjonesgoogle: It's not necessary for all fields to be in a flattening group. The point of a flattening group is to serve use cases which are both common and have a short list of parameters. Adding these two makes it a long list. I would just recommend not adding them; in Java, users can use the request-based method to access all parameters.

google/pubsub/v1/pubsub_gapic.yaml
garrettjonesgoogle: Seems like a bug in the config generator.

google/pubsub/v1/pubsub_gapic.yaml
garrettjonesgoogle: Yes, two messages will be generated. This isn't robust though, because if a third variant of the oneof is added which is of type Time or String, then we can't generate an overload for it. I think we need support for OneOf flattened parameters, as in https://github.com/googleapis/toolkit/issues/1057 . For the short term, just don't do any flattening at all.

google/pubsub/v1/pubsub_gapic.yaml
landrito: Simply removing the field name patterns.

Tracking this bug here: https://github.com/googleapis/toolkit/issues/1060

landrito:
@garrettjonesgoogle This is ready for another pass. PTAL.

google/pubsub/v1/pubsub_gapic.yaml
```diff
@@ -102,6 +104,19 @@ interfaces:
     field_name_patterns:
       subscription: subscription
     timeout_millis: 60000
+  - name: UpdateSubscription
+    flattening:
+      groups:
+      - parameters:
+        - subscription
+        - update_mask
```
garrettjonesgoogle: Due to usability concerns, I might not want to flatten this right now.

google/pubsub/v1/pubsub_gapic.yaml
```diff
@@ -102,6 +104,19 @@ interfaces:
...
+    required_fields:
+    - subscription
+    - update_mask
+    request_object_method: true
+    retry_codes_name: non_idempotent
```
garrettjonesgoogle: I think this should be idempotent conceptually.

google/pubsub/v1/pubsub_gapic.yaml
```diff
@@ -221,6 +236,64 @@ interfaces:
...
+      project: project
+    timeout_millis: 60000
+  - name: CreateSnapshot
+    flattening:
+      groups:
+      - parameters:
+        - name
+        - subscription
+    required_fields:
+    - subscription
```
garrettjonesgoogle: Like `CreateSubscription`, we probably want to include `name` here even though technically it isn't required.

google/pubsub/v1/pubsub_gapic.yaml
```diff
@@ -221,6 +236,64 @@ interfaces:
...
+    required_fields:
+    - snapshot
+    request_object_method: false
+    retry_codes_name: idempotent
+    retry_params_name: default
+    field_name_patterns:
+      snapshot: snapshot
+    timeout_millis: 60000
+  - name: Seek
+    # Removing flattening until we figure out how to deal with oneof
```
garrettjonesgoogle: It's odd to say that we're removing flattening when we never put it in... Just say that we aren't including it. Also include the issue #.

google/pubsub/v1/pubsub_gapic.yaml
landrito: Removed.

google/pubsub/v1/pubsub_gapic.yaml
landrito: Agreed. Fixed.

google/pubsub/v1/pubsub_gapic.yaml
landrito: Ah good catch.

google/pubsub/v1/pubsub_gapic.yaml
landrito: Done.

landrito:
@garrettjonesgoogle Ready for another pass. PTAL.

garrettjonesgoogle:
LGTM

PR #264, opened 2017-02-24
fix core config

pongad:


garrettjonesgoogle:
Adding @geigerj because he is the blamee for at least one of these fixes

PR #270, opened 2017-03-01
Allow a status code of UNAVAILABLE to be retried

jskeet:
This is applied to the `retry_codes` of every API.
It's only a change for non-idempotent methods; the idempotent
methods already included this in the list.


garrettjonesgoogle:
LGTM

PR #272, opened 2017-03-02
Remove RunQuery flattenings

chrisdunelm:
The flattening of RunQuery requires oneof support; remove the two flattenings that simulated oneof support using two overloads.

See https://github.com/googleapis/toolkit/issues/1057 for more details.

### DIFF
--- a/google/datastore/v1/datastore_gapic.yaml
+++ b/google/datastore/v1/datastore_gapic.yaml
@@ -56,6 +56,10 @@ interfaces:
     timeout_millis: 60000
   - name: RunQuery
     # TODO: Add flattening with oneof when oneofs implemented
+    required_fields:
+    - project_id
+    - partition_id
+    - read_options
     request_object_method: true
     retry_codes_name: idempotent
     retry_params_name: default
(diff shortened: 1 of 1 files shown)


chrisdunelm:
\cc @jskeet

jskeet:
For context:

- Java is still using the REST API for its library
- The C# code can (and will) add methods in a partial class to effectively put these overloads back, because they're useful and it's reasonable to predict that the oneof won't grow to have more fields. It's fine for another overload to be added later with the oneof though.

google/datastore/v1/datastore_gapic.yaml
```diff
@@ -55,22 +55,7 @@ interfaces:
...
-        - query
-      - parameters:
-        - project_id
-        - partition_id
-        - read_options
-        - gql_query
-    required_fields:
-    - project_id
-    - partition_id
-    - read_options
```
jskeet: Are they used for anything other than flattenings?

garrettjonesgoogle: Yes, they are used in languages that don't flatten.

chrisdunelm:
PTAL
required_fields removed by mistake, added back in.

garrettjonesgoogle:
LGTM

PR #273, opened 2017-03-02
Grpc bazel build support

makdharma:
WORKSPACE, BUILD and other supporting files needed to get the google.monitoring.v3.metric_service building with bazel.


bazel/cc_grpc_library.bzl
```diff
@@ -0,0 +1,61 @@
+"""Generates and compiles C++ grpc stubs from proto_library rules."""
```
ethanbao: What are the difference between bzl extension in this PR vs the one already in grpc github repo (https://github.com/grpc/grpc/blob/master/grpc.bzl)? And why these grpc bzl files part of googleapis not grpc?

google/protobuf/duration.proto
```diff
@@ -0,0 +1,104 @@
+// Protocol Buffers - Google's data interchange format
```
ethanbao: Adding protobuf protos into googleapis will cause confusion, as the same protos are already in https://github.com/google/protobuf/tree/master/src/google/protobuf

bazel/cc_grpc_library.bzl
makdharma: Got rid of the custom copy.

google/protobuf/duration.proto
makdharma: Unfortunately grpc bazel rules don't work with well known protos yet. Added a TODO to get rid of local copy when grpc supports this capability.

ethanbao: What are other options other than adding the protobuf protos here? Protos in this repo will be cloned/branched by developers, so we don't want to add temporary/incomplete protos here.

makdharma: The only other option (until first class support is available) is to create a shell script that does the copying of native protobuf files in preparation of running ```bazel build```.

makdharma: Never mind, we have a fix in the grpc tree. I'll modify this PR soon and remove local copy of google/protobuf.

ethanbao: Thanks for the fix.

makdharma:
@ethanbao PTAL

ethanbao:
LGTM

makdharma:
who needs to merge this PR?

qiwzhang:
This is bad.  This breaks many other project bazel builds.  We should not have any BUILD files in any of sub folders.

Other projects, such as Cloud endpoints,  are using their own BUILD files based on their own WORKSPACE dependency to build these proto files.

Here is the Could endpoint bazel BUILD file.

https://github.com/istio/proxy/blob/master/contrib/endpoints/repositories.bzl#L225

You can have BUILD file at the top, but not inside the subfolders.   When Endpoints BUILD file trying to use proto files inside the folder,  it if already has a BUILD file,  bazel will fail.

Can we roll this back?

makdharma:
No problem to roll it back. @ethanbao needs to do it?

nicolov:
I agree, I think you should roll back this change and have a single top-level BUILD file. Including this in a `new_http_archive` makes bazel read the nested BUILD files, and fail.

PR #279, opened 2017-03-03
Bump internal version to 0.15.3, GAX to 0.15.7

lukesneeringer:
@geigerj Hopefully another very boring PR for you.


lukesneeringer:
Note: Bumping to 0.15.3 because we have 0.15.2 for a couple individual libraries.

PR #281, opened 2017-03-06
Update logging bundling settings

michaelbausor:
Add flow control to logging bundling settings

This should not be merged until this toolkit branch is merged to master: https://github.com/googleapis/toolkit/tree/bundling-flow-control

### DIFF
--- a/google/logging/v2/logging_gapic.yaml
+++ b/google/logging/v2/logging_gapic.yaml
@@ -351,11 +351,11 @@ interfaces:
     resource_name_treatment: STATIC_TYPES
     bundling:
       thresholds:
-        element_count_threshold: 100
-        request_byte_threshold: 20480 # 20 Kb
-        delay_threshold_millis: 10
-        flow_control_element_limit: 10000
-        flow_control_byte_limit: 2048000 # 2000 Kb
+        element_count_threshold: 1000
+        request_byte_threshold: 1024000 # 1000 Kb
+        delay_threshold_millis: 50
+        flow_control_element_limit: 100000
+        flow_control_byte_limit: 10240000 # 10000 Kb
         flow_control_limit_exceeded_behavior: THROW_EXCEPTION
       bundle_descriptor:
         bundled_field: entries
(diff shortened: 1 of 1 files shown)


google/logging/v2/logging_gapic.yaml
```diff
@@ -352,8 +352,11 @@ interfaces:
     bundling:
       thresholds:
         element_count_threshold: 100
-        request_byte_threshold: 1024 # 1 Kb
+        request_byte_threshold: 20480 # 20 Kb
```
garrettjonesgoogle: What is the API limit for logging?

michaelbausor: The maximum size of a single entries write request is 5MB. So we could bump this way up (and the corresponding flow control setting).
https://cloud.google.com/logging/quota-policy

garrettjonesgoogle: I think it's really just the flow control limit that really matters for this. The request_byte_threshold value just means that the bundle will be sent. 

Let's set `flow_control_byte_limit` to 4MB for now because of https://github.com/GoogleCloudPlatform/google-cloud-java/issues/1499 , and add a comment saying that it can be bumped up to 5MB later.

michaelbausor: Why does this affect flow control and not `request_byte_threshold`? The 5MB limit is for a single request (i.e. a single bundle), which is what `request_byte_threshold` will affect. The flow control setting will affect the total number of bytes that can be queued in the bundler before additional calls either block or throw an exception. So the `request_byte_threshold` could be 4MB and the flow control setting could be 40MB, which would mean a max of approx 10 max size bundles could be queued.

garrettjonesgoogle: Ahh, right. It occurs to me that we don't actually have a place that can enforce message size limits. The bundling thresholds are to trigger bundling and don't throw errors if they are exceeded. 

But one thing we should at least do is not throw exceptions if we are capable of handling a larger message - e.g. grpc can handle 4MB, so the flow controller should be allowed to store that much at least.

michaelbausor:
Updated thresholds, PTAL

garrettjonesgoogle:
LGTM

PR #282, opened 2017-03-07
Change logging interface names

michaelbausor:


garrettjonesgoogle:
LGTM after you verify it looks right in api-client-staging

PR #283, opened 2017-03-08
vision: allow idempotent method to retry on INTERNAL error

pongad:
This reduces flakiness in Go tests.

Adding Vision POC @swcloud

PR #285, opened 2017-03-09
pubsub: match retry code with Java handwritten client

### LINKED ISSUE #284: pubsub's retry codes are not aggressive enough
The retry codes for publish should match [this method](https://github.com/GoogleCloudPlatform/google-cloud-java/blob/59459fc62bf649f8c1f4c3a0a6e625f5869e9b2a/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/Publisher.java#L484)

The retry codes for poll should match [this method](https://github.com/GoogleCloudPlatform/google-cloud-java/blob/59459fc62bf649f8c1f4c3a0a6e625f5869e9b2a/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/spi/v1/StatusUtil.java#L27)

pongad:
Fixes #284.

### DIFF
--- a/google/pubsub/v1/pubsub_gapic.yaml
+++ b/google/pubsub/v1/pubsub_gapic.yaml
@@ -47,7 +47,14 @@ interfaces:
     - DEADLINE_EXCEEDED
   - name: non_idempotent
     retry_codes:
-    - UNAVAILABLE      
+    - UNAVAILABLE
+  - name: poll
+    retry_codes:
+    - DEADLINE_EXCEEDED
+    - INTERNAL
+    - CANCELLED
+    - RESOURCE_EXHAUSTED
+    - UNAVAILABLE
   retry_params_def:
   - name: default
     initial_retry_delay_millis: 100
@@ -194,7 +201,7 @@ interfaces:
     - max_messages
     request_object_method: true
     resource_name_treatment: STATIC_TYPES
-    retry_codes_name: non_idempotent
+    retry_codes_name: poll
     retry_params_name: messaging
     field_name_patterns:
       subscription: subscription
@@ -207,7 +214,7 @@ interfaces:
       response:
         resources_field: received_messages
     resource_name_treatment: STATIC_TYPES
-    retry_codes_name: non_idempotent
+    retry_codes_name: poll
     retry_params_name: messaging
     field_name_patterns:
       subscription: subscription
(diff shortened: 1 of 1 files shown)


google/pubsub/v1/pubsub_gapic.yaml
```diff
@@ -47,7 +47,14 @@ interfaces:
     - DEADLINE_EXCEEDED
   - name: non_idempotent
     retry_codes:
-    - UNAVAILABLE      
+    - UNAVAILABLE
+  - name: poll
```
garrettjonesgoogle: This should be `pull`, not `poll`

pongad: You're right. Done

pongad:
@garrettjonesgoogle PTAL

google/pubsub/v1/pubsub_gapic.yaml
```diff
@@ -374,11 +381,16 @@ interfaces:
     - DEADLINE_EXCEEDED
   - name: one_plus_delivery
     retry_codes:
+    - ABORTED
+    - CANCELLED
+    - INTERNAL
```
garrettjonesgoogle: The comments by @jba on https://github.com/googleapis/googleapis/pull/283 concern me about retrying `INTERNAL`.

pongad: Hmm, this is a good point. @davidtorres ?

davidtorres: @pongad , INTERNAL is certainly a complicated status to handle, as @jba clearly points out,  internal errors must be caused by invariants which usually surface by inputs that were not expected by the system and will permanently fail; a bug in the system. However we often see *internal* errors that surface just because an unexpected failure coming from an upstream dependency or an environment issue that usually gets resolved by itself; a transient error, such errors should genuinely be reported as unavailable errors but given our quite complex infrastructure, that is not always the case. Hence, in the pubsub library we have opted to a more conservative approach of retrying even on internal errors.

garrettjonesgoogle: @davidtorres thanks for the useful background!

garrettjonesgoogle:
LGTM

PR #290, opened 2017-03-16
pubsub java: deprecate Pull, ModifyAckDeadline, and Acknowledge

pongad:
These functionalities are instead exposed in Subscriber class.

Updates GoogleCloudPlatform/google-cloud-java#1734 .


garrettjonesgoogle:
LGTM

PR #298, opened 2017-03-30
Remove google/api/BUILD and google/rpc/BUILD

ethanbao:
They are currently blocking other repos to use this repo as bazel repo dependency.


ethanbao:
@qiwzhang @makdharma

PR #302, opened 2017-03-31
Add package_type for all APIs and iam version for Java

shinfan:


gapic/packaging/dependencies.yaml
```diff
@@ -91,6 +91,9 @@ google-iam-v1_version:
   ruby:
     name_override: grpc-google-iam
     lower: '0.6.8'
+  java:
+    name_override: grpc-google-iam-v1
```
garrettjonesgoogle: Why do we need a name override?

shinfan: I think the default name is google-iam-v1, python does the same override.

garrettjonesgoogle: But for Java, we just always prepend "grpc-" to every package, which seems like we could do in the transformer without having to do any name overrides in the config.

shinfan: I agree that currently it is feasible for Java to do it in the transformer. But it looks risky to me that we assume it is always correct for Java to append 'grpc-' to the default name. This introduces the extra dependency of the default name in gapic. I feel we should follow the existing overriding method here as other languages.

garrettjonesgoogle: Why would it be risky? If we have to override for every package, then every new package requires this manual work, and people could make mistakes with it. You need to add this same override for every service, right?

shinfan: Fair enough. I have moved this into toolkit transformer.

shinfan: Toolkit change: https://github.com/googleapis/toolkit/pull/1159/commits/b4cd5ac97937a779a349caa63028e0358f8e9d9f

gapic/api/artman_bigtable.yaml
```diff
@@ -1,6 +1,7 @@
 common:
   api_name: bigtable
   api_version: v2
+  package_type: grpc_client
```
geigerj: Given that the vast majority of artman configs are likely to be gRPC clients, can this just be a default setting, so that the field only needs to be configured if it's *not* a gRPC client?

shinfan: I have thought about using default but it seems like we do not have such a default yaml (it seems overkill to create a default config file just for this one field).
On the other hand, does Artman tasks allows default values for a field? If I do not add those fields for every API, Artman task just gives me a runtime exception of missing requirements.

geigerj: Right... Is it possible to use [common.yaml](https://github.com/googleapis/googleapis/blob/master/gapic/lang/common.yaml) to store default values? I think that would be a better option than implementing the logic in `artman`, since you'd have to change the config parsing code to avoid the runtime exception. WDYT?

shinfan: Unfortunately the common.yaml won't work in this case. Based on the config loading logic, the config is loaded in sequence which means the later loaded ones will override the previous ones. Since this PR is blocking Java GA I suggest we tackle this problem in a follow-up PR. Issue created: https://github.com/googleapis/artman/issues/148

garrettjonesgoogle:
LGTM

PR #304, opened 2017-04-03
Switch timeout_mills to 190000 for Speech v1

shinfan:
More context: https://github.com/GoogleCloudPlatform/google-cloud-ruby/pull/1380

cc/ @pbogle

### DIFF
--- a/google/cloud/speech/v1/cloud_speech_gapic.yaml
+++ b/google/cloud/speech/v1/cloud_speech_gapic.yaml
@@ -56,7 +56,7 @@ interfaces:
     request_object_method: true
     retry_codes_name: idempotent
     retry_params_name: default
-    timeout_millis: 60000
+    timeout_millis: 190000
   - name: LongRunningRecognize
     flattening:
       groups:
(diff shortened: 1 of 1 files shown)


google/cloud/speech/v1/cloud_speech_gapic.yaml
```diff
@@ -74,7 +74,7 @@ interfaces:
     request_object_method: true
     retry_codes_name: non_idempotent
     retry_params_name: default
-    timeout_millis: 60000
+    timeout_millis: 190000
```
pbogle: Technically, LongRunningRecognize doesn't need to be this long since we're just waiting for the LRO Name to be returned, not the actual recognition.  30 seconds would be more than long enough for the worst possible cases.  This probably isn't a critical issue though, is there any harm to making this too long?

lukesneeringer:
Oh, there was another one that needed to be 190? Rats.
All the client libraries will need to be regenerated (or manually edited).

shinfan:
PTAL

lukesneeringer:
@landrito @shinfan I am not sure if this value needs to be changed. There is likely no risk though. I would like to know if the Ruby folks just had an outdated GAPIC, and my change in #297 would have fixed them.

google/cloud/speech/v1/cloud_speech_gapic.yaml
shinfan: Good point. LRO should not need this long. Reverted.

PR #306, opened 2017-04-04
Made devtools/source/v1 visible to grpc generator

landrito:
Looks like the grpc generator is missing the source protos without this.

https://github.com/googleapis/api-client-staging/pull/207#issuecomment-291586734


geigerj:
@landrito should other language leads be aware of this, since it'll require regenerating their gRPC libraries?

landrito:
Good point. I think only @pongad needs to be made aware since only go and ruby have been generated to api-client-staging.

PR #308, opened 2017-04-06
(DO NOT MERGE) Config+proto for nonfunctional compute client

garrettjonesgoogle:
This isn't for review exactly, it's more to look at for amusement.


google/compute/compute.yaml
```diff
@@ -0,0 +1,23 @@
...
+    Google Compute Engine Service
+
+apis:
+- name: google.compute.v1.InstancesService
+
+authentication:
+  rules:
+    - selector: '*'
+      oauth:
+        canonical_scopes: |-
```
bjwatson: Looks like every method in the Discovery doc has its own scopes. That makes things interesting. As a simple starting point, we could simply union them all to generate this.

google/compute/v1/compute_gapic.yaml
```diff
@@ -0,0 +1,101 @@
...
+    retry_codes_name: idempotent
+    retry_params_name: default
+    timeout_millis: 60000
+  - name: InsertInstance
+    flattening:
+      groups:
+      - parameters:
+        - project
+        - zone
+        - resource
```
bjwatson: It seems better to call this `instance`, based on 

```json
     "request": {
      "$ref": "Instance"
     },
```

google/compute/compute.yaml
garrettjonesgoogle: That's a good starting point... maybe it's the ending point too :-)

google/compute/v1/compute_gapic.yaml
garrettjonesgoogle: (This comment thread would probably be better on the proto)

The discovery doc doesn't have a name for this field because it is just set as the post body. The original proto2 called it `resource`. Maybe `instance` is a better name as you say.

lukesneeringer:
@garrettjonesgoogle What is the plan for this? (Just wanting to do some general low-hanging cleanup on the "easy" repos.)

bjwatson:
@lukesneeringer This was an experimental branch for Discovery codegen. Up to @garrettjonesgoogle and @andreamlin whether this is still needed.

garrettjonesgoogle:
@lukesneeringer it's not strictly needed, but it might be nice to have it around for another couple months.

Issue #303, opened 2017-04-02
API Key Restrictions for iOS Apps with Bundle Identifier


imeier:
When calling https://servicecontrol.googleapis.com/v1/services/{serviceName}:check, how is the bundle identifier supposed to be passed so that an API key can be validated for an iOS app restricted to a bundle identifier?

Thanks,
Ian

imeier:
This issue has been resolved: https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!topic/google-cloud-endpoints/I-u3sAUU3Ts.

However, the documentation should be updated to reflect this.

Thanks,
Ian

PR #309, opened 2017-04-11
Updating Service Control Operation Label Documentation for Restricted API Keys

imeier:
This pull request documents the required service control operation labels to use when validating restricted API keys via the check method, as documented here: https://github.com/googleapis/googleapis/issues/303. The added documentation is based on the https://github.com/cloudendpoints/endpoints-management-python project.

The documentation should also be updated here: https://cloud.google.com/service-control/reference/rest/v1/Operation.

Thanks,
Ian


geigerj:
Hi @imeier -- thank you for the contribution! Unfortunately, we cannot accept PRs to the proto files contained in the `google/` directory of this repository, as they are automatically synchronized by a one-way process from a Google-internal codebase; any commits to this proto will be overwritten when it's next synchronized. However, I've relayed your PR to the team that owns this proto.

PR #310, opened 2017-04-11
prepare release 0.1.7

pongad:
grpc and protobuf bumped to 1.2.0 and 3.2.0 respectively

Can we roll this out before gax and google-cloud-java? Or do we need to big bang?


garrettjonesgoogle:
It looks like we don't have any imminent API updates, so let's commit this now. First, update the commit message to "Prepare generated grpc package release 0.1.7 for Java", then LGTM to submit.

PR #312, opened 2017-04-13
Cut polling interval to 20000 (1/3) for Speech v1 LRO

shinfan:
Long term speaking we should implement the exponential back-off to fix the issue, but this should be good for now.


garrettjonesgoogle:
Confirm the value with the API owners.

PR #314, opened 2017-04-17
Update Java dependency and generated package versions

michaelbausor:


garrettjonesgoogle:
LGTM, but change the commit message - suggestion: "Update Java dependency and generated package versions"

PR #316, opened 2017-04-18
Expanded artman_*.yaml files, compatible with Artman 0.2.0

lukesneeringer:
This should get _most_ configuration in place for googleapis/artman#159.

I was not sure what to do about C#, since I had no staged API clients to work against, but I have reached out to Jon Skeet. Using `--publish=noop` will retain the old behavior. The rest should be mostly correct but may need some slight tinkering by the relevant language leads.


lukesneeringer:
This was approved by @michaelbausor this morning in Hangouts. As artman 0.2.0 has gone in, merging this.

PR #319, opened 2017-04-20
Update Artman YAMLs.

lukesneeringer:
Companion to googleapis/artman#162.

### DIFF
--- a/gapic/api/artman_bigtable.yaml
+++ b/gapic/api/artman_bigtable.yaml
@@ -6,46 +6,65 @@ common:
   proto_deps:
     - google-common-protos
   import_proto_path:
-    - ${REPOROOT}/googleapis
+    - ${GOOGLEAPIS}
   src_proto_path:
-    - ${REPOROOT}/googleapis/google/bigtable/v2
+    - ${GOOGLEAPIS}/google/bigtable/v2
   service_yaml:
-    - ${REPOROOT}/googleapis/google/bigtable/bigtable.yaml
+    - ${GOOGLEAPIS}/google/bigtable/bigtable.yaml
   output_dir: ${REPOROOT}/artman/output
+  git_repos:
+    staging:
+      location: git@github.com:googleapis/api-client-staging.git
   gapic_api_yaml:
-    - ${REPOROOT}/googleapis/google/bigtable/v2/bigtable_gapic.yaml
+    - ${GOOGLEAPIS}/google/bigtable/v2/bigtable_gapic.yaml
 csharp:
   gapic_code_dir: ${REPOROOT}/artman/output/gcloud-csharp-cloud-bigtable
 go:
   gapic_code_dir: ${REPOROOT}/gapi-cloud-bigtable-go
-  git_repo:
-    location: git@github.com:GoogleCloudPlatform/google-cloud-go.git
-    gapic_subpath: bigtable/apiv2
+  git_repos:
+    go:
+      location: git@github.com:GoogleCloudPlatform/google-cloud-go.git
+      gapic_subpath: bigtable/apiv2
+    staging:
+      gapic_subpath: google-cloud-bigtable-v2/cloud.google.com/go/bigtable/apiv2
 java:
-  gapic_code_dir: ${REPOROOT}/google-cloud-java/google-cloud-bigtable
-  git_repo:
-    gapic_subpath: generated/java/google-cloud-bigtable-v2
+  gapic_code_dir: ${REPOROOT}/artman/output/java/google-cloud-bigtable
+  git_repos:
+    java:
+      gapic_component: src/main/java/com/google/cloud/bigtable/spi/v2
+      gapic_subpath: google-cloud-pubsub/src/main/java/com/google/cloud/bigtable/spi/v2
+      location: git@github.com:GoogleCloudPlatform/google-cloud-java.git
+    staging:
+      gapic_subpath: generated/java/google-cloud-bigtable-v2
 nodejs:
   gapic_code_dir: ${REPOROOT}/artman/output/js/bigtable-v2
-  git_repo:
-    location: git@github.com:GoogleCloudPlatform/google-cloud-node.git
-    gapic_component: src/v2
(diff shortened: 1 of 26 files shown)


gapic/api/artman_bigtable.yaml
```diff
@@ -6,46 +6,65 @@ common:
...
+      gapic_subpath: google-cloud-bigtable-v2/cloud.google.com/go/bigtable/apiv2
 java:
-  gapic_code_dir: ${REPOROOT}/google-cloud-java/google-cloud-bigtable
-  git_repo:
-    gapic_subpath: generated/java/google-cloud-bigtable-v2
+  gapic_code_dir: ${REPOROOT}/artman/output/java/google-cloud-bigtable
+  git_repos:
+    java:
+      gapic_component: src/main/java/com/google/cloud/bigtable/spi/v2
+      gapic_subpath: google-cloud-pubsub/src/main/java/com/google/cloud/bigtable/spi/v2
```
garrettjonesgoogle: * The value that I see here in `gapic_component` is generated inside toolkit for java; this doesn't need to be set at all.
* `gapic_subpath` should just be `google-cloud-pubsub`

gapic/api/artman_bigtable.yaml
```diff
@@ -6,46 +6,65 @@ common:
...
-  git_repo:
-    location: git@github.com:GoogleCloudPlatform/google-cloud-ruby.git
-    gapic_component: lib/google/cloud/bigtable/v2
-    gapic_subpath: google-cloud-bigtable/lib/google/cloud/bigtable/v2
+  gapic_code_dir: ${REPOROOT}/artman/output/ruby/google-cloud-bigtable
+  git_repos:
+    ruby:
+      location: git@github.com:GoogleCloudPlatform/google-cloud-ruby.git
+      gapic_component: lib/google/cloud/bigtable/v2
+      gapic_subpath: google-cloud-bigtable/lib/google/cloud/bigtable/v2
```
garrettjonesgoogle: Why the redundancy between `gapic_component` and `gapic_subpath`?

lukesneeringer: How this is used conceptually is "cp `{generated code}/{component} -rf {destination repo}/{subpath}".

Even though a lot of it looks duplicative, there is an important distinction. In Node and Ruby, the GAPIC gets "tucked" inside of an existing manual library. Since this process involves removing everything at `subpath` before copying the new components, this staging process would effectively delete the manual layer completely.

Additionally, in Node and Ruby, only a subset of the generated code is used in these cases (the package metadata and outer `index` files are thrown away). However, the entire generated code is used in GAPIC only (so the settings would be distinct for a GAPIC only library).

Basically: this is not _actually_ redundant; there is a specific semantic meaning, which is to copy only this slice of the GAPIC into this particular slice of the target repo.

gapic/api/artman_bigtable.yaml
lukesneeringer: Okay, I got that wrong for Java. :-) Quick question: Would you check on the natural language ones and tell me what the correct values are there? (See my explanation of the values in your other comment for details.) With the settings you are asking for, the side-by-side case would end up stomping over itself.

gapic/api/artman_language_v1.yaml
```diff
@@ -6,46 +6,65 @@ common:
...
+      gapic_subpath: google-cloud-language-v1/cloud.google.com/go/language/apiv1
 java:
-  gapic_code_dir: ${REPOROOT}/google-cloud-java/google-cloud-language
-  git_repo:
-    gapic_subpath: generated/java/google-cloud-language-v1
+  gapic_code_dir: ${REPOROOT}/artman/output/java/google-cloud-language
+  git_repos:
+    java:
+      gapic_component: src/main/java/com/google/cloud/language/spi/v1
+      gapic_subpath: google-cloud-pubsub/src/main/java/com/google/cloud/language/spi/v1
```
garrettjonesgoogle: This should be `google-cloud-language` (it looks like you have `google-cloud-pubsub` in a number of places)

gapic/api/artman_bigtable.yaml
garrettjonesgoogle: The side-by-side case will still work fine for Java.

* In the google-cloud repo, the existing code is not removed before new code is copied in, so there is no problem
* In the staging repo, the existing code is removed, but the roots are distinct, so they won't stomp over each other

gapic/api/artman_language_v1.yaml
lukesneeringer: Thanks. (I used pubsub as my copy / paste template. Eek!)

gapic/api/artman_bigtable.yaml
lukesneeringer: > In the google-cloud repo, the existing code is not removed before new code is copied in, so there is no problem

`--staging=local` very much removes what used to be at that spot ([here][1]). This is also true of the Artman [_status quo ante_][2] before my refactor last week.

  [1]: https://github.com/googleapis/artman/blob/master/artman/tasks/publish/local.py#L59
  [2]: https://github.com/googleapis/artman/blob/e45376ffd4eb6d3d9f9f3bdac5706cea9adf4d70/pipeline/tasks/gapic_tasks.py#L120

That said, my feeling on the language specific segments of these files is that the language owners decide. Therefore, I will have this changed to what you are asking momentarily. If it turns out it is not right for you, you are always free to change it (and you very much do not need my buy-in to do so). :-)

garrettjonesgoogle: Maybe we have a different understanding of staging? For me, it is 1:1 with api-client-staging, and putting code into google-cloud-java isn't called "staging".

gapic/api/artman_language_v1.yaml
lukesneeringer: Just grepped every non-pubsub file to ensure I got them all. Thank you!

lukesneeringer:
@garrettjonesgoogle I believe I updated all the Java YAML segments to what you are expecting. Let me know if what is there looks right for your workflow.

PR #321, opened 2017-04-26
Add Python beta client library version configuration

geigerj:
See also: https://github.com/googleapis/toolkit/pull/1216, https://github.com/googleapis/artman/pull/171


geigerj:
Note: this PR only updates the configuration for Python. I'll do the other languages as a follow-up PR.

PR #322, opened 2017-04-26
make errorreporting smoke test work

pongad:
The location of the error must be encoded either in `event.message` or `event.context`. This PR make the smoke test not return an invalid argument error.

### DIFF
--- a/google/devtools/clouderrorreporting/v1beta1/errorreporting_gapic.yaml
+++ b/google/devtools/clouderrorreporting/v1beta1/errorreporting_gapic.yaml
@@ -35,7 +35,7 @@ interfaces:
     - DEADLINE_EXCEEDED
   - name: non_idempotent
     retry_codes:
-    - UNAVAILABLE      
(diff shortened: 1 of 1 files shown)


google/devtools/clouderrorreporting/v1beta1/errorreporting_gapic.yaml
```diff
@@ -80,6 +80,10 @@ interfaces:
     init_fields:
     - project_name%project=$PROJECT_ID
     - event.message="[MESSAGE]"
+    - event.service_context.service="[SERVICE]"
+    - event.context.report_location.file_path="path/to/file.lang"
```
shinfan: Do we need to setup such a file for each language to run the smoke test?

pongad: No, I think you just need the fields set. It can just be the same for all languages.

shinfan:
LGTM

PR #329, opened 2017-04-27
Bump the beta package versions for Python

lukesneeringer:


gapic/api/artman_datastore.yaml
```diff
@@ -51,10 +51,10 @@ php:
     staging:
       gapic_subpath: generated/php/google-cloud-datastore-v1
   enable_batch_generation: True
-python:
```
lukesneeringer: LOL @ random extra space that my editor (correctly) stripped.

PR #332, opened 2017-04-28
Support updated batch pipelines

michaelbausor:
Paired with https://github.com/googleapis/artman/pull/179

### DIFF
--- a/gapic/batch/common.yaml
+++ b/gapic/batch/common.yaml
@@ -2,14 +2,3 @@ common:
   batch_apis: '*'
   api_config_pattern: ${GOOGLEAPIS}/gapic/api/artman_${API_SHORT_NAME}.yaml
   artman_language_yaml: ${GOOGLEAPIS}/gapic/lang/common.yaml
-  git_repos:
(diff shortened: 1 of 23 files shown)


gapic/batch/common.yaml
```diff
@@ -1,5 +1,15 @@
 common:
   batch_apis: '*'
-  batch_languages: '*'
   api_config_pattern: ${GOOGLEAPIS}/gapic/api/artman_${API_SHORT_NAME}.yaml
   artman_language_yaml: ${GOOGLEAPIS}/gapic/lang/common.yaml
+  git_repos:
```
geigerj: Why is this empty configuration necessary at all?

michaelbausor: This is because artman currently requires the `git_repos` arg - but it is better handled in artman itself, rather than with empty configuration. Removed.

michaelbausor:
@lukesneeringer I'm not sure I understand - longrunning as in the GAPIC feature? That isn't affected by batch generation.

michaelbausor:
Figured it out, yes you are right - disabled.

PR #335, opened 2017-05-02
Convert artman_* files to new format.

lukesneeringer:
Companion PR to googleapis/artman#181.

### DIFF
--- a/gapic/api/artman_bigtable.yaml
+++ b/gapic/api/artman_bigtable.yaml
@@ -24,46 +24,57 @@ go:
   git_repos:
     go:
       location: git@github.com:GoogleCloudPlatform/google-cloud-go.git
-      gapic_subpath: bigtable/apiv2
+      paths:
+        - bigtable/apiv2
     staging:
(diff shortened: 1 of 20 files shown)


gapic/api/artman_bigtable.yaml
```diff
@@ -24,46 +24,57 @@ go:
...
+      paths:
+        - generated/go/google-cloud-bigtable-v2
 java:
   gapic_code_dir: ${REPOROOT}/artman/output/java/google-cloud-bigtable
   git_repos:
     java:
       location: git@github.com:GoogleCloudPlatform/google-cloud-java.git
-      gapic_subpath: google-cloud-bigtable
+      paths:
+        - google-cloud-bigtable
```
ethanbao: nit: indentation. Not sure whether yaml is indentation sensitive.

lukesneeringer: It is indentation sensitive (very much so), but the indentation I am using here is correct.

PR #339, opened 2017-05-10
Add Java staging paths

shinfan:


shinfan:
Friendly ping

michaelbausor:
LGTM - it would also be nice to do some of the other APIs, even if it is not necessary yet (such as cloud debugger, datastore, bigtable, functions)

shinfan:
@michaelbausor I feel it is confusing to setup staging paths for APIs that do not exist in the java staging repo. We can add the paths on demand if we need in the future.

michaelbausor:
Unfortunately they already have paths, just the wrong ones after the change: https://github.com/googleapis/googleapis/blob/master/gapic/api/artman_datastore.yaml#L41

shinfan:
Oh that's a different story then. I will update these paths as well.

PR #344, opened 2017-05-18
Video Intelligence is public now.

lukesneeringer:
So we do not need to be committing to `*-private` anymore. :-)


ethanbao:
LGTM

PR #346, opened 2017-05-19
Update capitalization in PHP video intelligence namespace

michaelbausor:


garrettjonesgoogle:
Is the config generator still generating capital `B`?

michaelbausor:
Config gen is not in a great place - for video intelligence, the only language it gets completely right is Java. For C#, Ruby, PHP it produces `Videointelligence` instead of `VideoIntelligence`, for python it misses `gapic`, for go it adds an extra `cloud`, and for python and node it omits the domain layer location.

However, it does produce `V1beta1` with a lower case `b`

garrettjonesgoogle:
LGTM

PR #349, opened 2017-05-20
Add core path to the batch generation

shinfan:
This PR enables both core and staged apis will be generated by the batch command.

Artman PR: https://github.com/googleapis/artman/pull/194


michaelbausor:
LGTM once https://github.com/googleapis/artman/pull/194 is approved

PR #352, opened 2017-05-23
NodeJS: Change output paths for node to match api-client-staging

landrito:


landrito:
Verified that this will fix generation problems with video intelligence.

PR #362, opened 2017-05-31
Update config for DLP

swcloud:


swcloud:
@geigerj Do we still manually add config for LRO in gapic config yaml?

neozwu:
It seems speech and videointelligence don't have manual config for LRO.
/cc @shinfan

shinfan:
If it is not generated by the config gen then yes we probably still need to add it manually. 
Have created an issue to track this: https://github.com/googleapis/toolkit/issues/1321

PR #363, opened 2017-06-01
Add Ruby DLP configuration

geigerj:


gapic/api/artman_dlp.yaml
```diff
@@ -49,3 +49,15 @@ python:
     staging:
       gapic_subpath: generated/python/gapic-google-cloud-dlp-v2beta1
       grpc_subpath: generated/python/proto-google-cloud-dlp-v2beta1
+ruby:
+  gapic_code_dir: ${REPOROOT}/artman/output/ruby/google-cloud-dlp
+  git_repos:
+    ruby:
+      paths:
+        - src: lib/google/cloud/dlp/v1
+          dest: google-cloud-dlp/lib/google/cloud/dlp/v1
```
swcloud: Do you need to set up the 'test' path mapping for unit tests?

geigerj: Ah, I'm not familiar with the test path mapping. I followed the pattern for Ruby PubSub here. Could you point me to an example of how the test path should be configured?

swcloud: https://github.com/googleapis/googleapis/blob/master/gapic/api/artman_dlp.yaml#L42

lukesneeringer: @geigerj Use Vision or Video as the reference.

geigerj: Done.

geigerj:
@swcloud Could you take a look at this one, too?

PR #365, opened 2017-06-03
Add test dependency for pubsub

shinfan:

### DIFF
--- a/gapic/packaging/dependencies.yaml
+++ b/gapic/packaging/dependencies.yaml
@@ -49,7 +49,7 @@ gax_version:
   ruby:
     lower: '0.8.0'
   java:
-    lower: '1.2.0'
+    lower: '0.6.1'
 
 proto_version:
   python:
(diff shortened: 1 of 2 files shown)


gapic/packaging/dependencies.yaml
```diff
@@ -49,7 +49,11 @@ gax_version:
   ruby:
     lower: '0.8.0'
   java:
-    lower: '0.6.1'
+    lower: '1.2.0'
+
+gax_grpc_version:
```
garrettjonesgoogle: Do you have to make a code change to use this?

shinfan: Yes, I have one PR pending that will use this. This should not break the existing code though. I can revert this for now for safety.

shinfan:
@garrettjonesgoogle PTAL

gapic/packaging/dependencies.yaml
garrettjonesgoogle: Hmm well gax 1.2.0 is incomplete, so someone using only this shouldn't be able to build a coherent package, right?

shinfan: ahhh okay I have reverted the whole file and add the change later once my code PR is complete.

garrettjonesgoogle:
LGTM

PR #372, opened 2017-06-08
Update publishing paths for Python.

lukesneeringer:


gapic/api/artman_speech_v1beta1.yaml
```diff
@@ -32,6 +32,16 @@ go:
 java:
   gapic_code_dir: ${REPOROOT}/artman/output/java/google-cloud-speech
   git_repos:
+    python:
```
geigerj: Why is this under the `java` section?

lukesneeringer: Because I derped.

PR #373, opened 2017-06-09
logging: retry on INTERNAL error

pongad:
Logging team communicated that this is OK.

@jba:
Want to make sure, since the email you forwareded me didn't call this out explicitly. Retrying on internal is OK even if methods are not idempotent right?


jba:
I would only do it for idempotent methods.

pongad:
@jba PTAL

PR #383, opened 2017-06-20
Adjust C# output directories to be structured like Java.

jskeet:


gapic/api/artman_pubsub.yaml
```diff
@@ -24,7 +24,7 @@ common:
     staging:
       location: git@github.com:googleapis/api-client-staging.git
 csharp:
-  gapic_code_dir: ${REPOROOT}/artman/output/gcloud-csharp-pubsub
+  gapic_code_dir: ${REPOROOT}/artman/csharp/java/google-cloud-pubsub
```
ethanbao: csharp/java?

gapic/api/artman_longrunning.yaml
```diff
@@ -19,7 +19,7 @@ common:
     staging:
       location: git@github.com:googleapis/api-client-staging.git
 csharp:
-  gapic_code_dir: ${REPOROOT}/artman/output/gax-dotnet
+  gapic_code_dir: ${REPOROOT}/artman/output/csharp/google-longrunning
```
ethanbao: To confirm this is intended, right?

gapic/api/artman_pubsub.yaml
jskeet: Humbug - will fix, thanks.

gapic/api/artman_longrunning.yaml
jskeet: Correct. LRO isn't in GAX for .NET - it's a separate package.

gapic/api/artman_pubsub.yaml
jskeet: (Done and amended the commit)

PR #390, opened 2017-06-29
use resource name type for DLP generation

neozwu:


garrettjonesgoogle:
Regenerate your api-client-staging PR with this so I can see the effect on the generated client

PR #394, opened 2017-06-30
toolkit requires nodejs package name to be lower_underscore

swcloud:


landrito:
I have a hunch that this should be solved in toolkit.

landrito:
Closing since this was solved in toolkit.

PR #397, opened 2017-07-06
add smoke test for DLP

neozwu:
Include a smoke test config in gapic_yaml


neozwu:
@pongad No, it's a randomly-generated-by-human number. Obviously I'm not a good random number generator :)

PR #402, opened 2017-07-12
Fix PHP Namespace casing in video intelligence proto

bshaffer:
Right now the protobuf library is generating the following namespace:

```php
namespace Google\Cloud\Videointelligence\V1beta1;
```

But it should be:

```php
namespace Google\Cloud\VideoIntelligence\V1beta1;
```

Let's take advantage of the `php_namespace` option so that this will be consistent with the other libraries.


bshaffer:
This was fixed upstream.

PR #403, opened 2017-07-13
Fix the streaming defaults

lukesneeringer:
The current default values cause the stream to be aborted and retried every 12 seconds. This also causes Pub/Sub to drop lease delays. The `streaming_pull` method should have a much, much longer keepalive.


garrettjonesgoogle:
LGTM

PR #411, opened 2017-07-24
Bump Publish, StreamingPull maximum timeouts

geigerj:
During benchmark testing, @blowmage observed a long retry loop
due to continual timeouts when multiple Publishers are used on
and limited-bandwidth connection. This increases the maximum
timeout to address the specific circumstances under which this
behavior was observed. (See
https://github.com/googleapis/gax-ruby/issues/79)

Further, some gRPC languages kill the stream after the timeout;
this affects StreamingPull, where with current configuration, the
Ruby stream will stay open only 12s. Instead bump this to the
maximum 10 minutes.

Updates https://github.com/googleapis/gax-ruby/issues/79

### DIFF
--- a/google/pubsub/v1/pubsub_gapic.yaml
+++ b/google/pubsub/v1/pubsub_gapic.yaml
@@ -74,6 +74,14 @@ interfaces:
     rpc_timeout_multiplier: 1
     max_rpc_timeout_millis: 12000 # 12 seconds
     total_timeout_millis: 600000 # 10 minutes
+  - name: streaming_messaging
+    initial_retry_delay_millis: 100
+    retry_delay_multiplier: 1.3
+    max_retry_delay_millis: 60000 # 60 seconds
+    initial_rpc_timeout_millis: 600000 # 10 minutes
+    rpc_timeout_multiplier: 1
+    max_rpc_timeout_millis: 600000 # 10 minutes
+    total_timeout_millis: 600000 # 10 minutes
   experimental_features:
     iam_resources:
     - type: google.pubsub.v1.Subscription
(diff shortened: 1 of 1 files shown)


geigerj:
ping @lukesneeringer

google/pubsub/v1/pubsub_gapic.yaml
```diff
@@ -434,7 +434,7 @@ interfaces:
     max_retry_delay_millis: 60000 # 60 seconds
     initial_rpc_timeout_millis: 12000 # 12 seconds
     rpc_timeout_multiplier: 1
-    max_rpc_timeout_millis: 12000 # 12 seconds
+    max_rpc_timeout_millis: 30000 # 30 seconds
```
blowmage: This means that GAPIC will be killing the stream every 30 seconds. Can we get this closer to the 10 minutes total timeout?

Do we have an existing issue to not kill long-living streams?

geigerj: This config is only used for the `publish` method on the Publisher. Subscriber StreamingPull uses https://github.com/googleapis/googleapis/pull/411/files#diff-1445f68867b94d222bea9502a59f7c39R69. If I recall correctly from an internal discussion a while ago, we also wanted to change that; let me amend this PR.

blowmage: Gotcha. Apologies for my confusion. (I realized that last week, but forgot about it when I looked at this today.)

google/pubsub/v1/pubsub_gapic.yaml
```diff
@@ -74,6 +74,14 @@ interfaces:
     rpc_timeout_multiplier: 1
     max_rpc_timeout_millis: 12000 # 12 seconds
     total_timeout_millis: 600000 # 10 minutes
+  - name: streaming_messaging
+    initial_retry_delay_millis: 100
+    retry_delay_multiplier: 1.3
+    max_retry_delay_millis: 60000 # 60 seconds
+    initial_rpc_timeout_millis: 600000 # 10 minutes
+    rpc_timeout_multiplier: 1
+    max_rpc_timeout_millis: 600000 # 10 minutes
```
blowmage: :+1:

PR #422, opened 2017-08-16
spanner: enable batch in go

pongad:
Previously Go's handwritten client uses the raw grpc client.
We're migrating the handwritten client to use GAPIC.


pongad:
@garrettjonesgoogle Can you take a look since Jonathan is OOO?

PR #427, opened 2017-08-18
Restructure api_defaults to allow multiple versions

shinfan:
Tookit PR: https://github.com/googleapis/toolkit/pull/1491
Artman PR: https://github.com/googleapis/artman/pull/255

### DIFF
--- a/gapic/packaging/api_defaults.yaml
+++ b/gapic/packaging/api_defaults.yaml
@@ -23,22 +23,22 @@ release_level:
   python: alpha
   java: alpha
 generated_package_version:
-  non_ga:
-    # TODO: Python must go to 1.0.20 or 1.1.0 when we GA, due to mistakenly
-    #       publishing some 1.0.x versions to pypi already. They are now
-    #       deleted but files still exist on the pypi servers and can prevent
-    #       us from using the version number again.
-    python:
-      lower: '0.15.4'
-      upper: '0.16dev'
-    nodejs:
-      lower: '0.7.1'
-    php:
-      lower: '0.1.0'
-    ruby:
-      lower: '0.6.8'
-    java:
-      lower: '0.1.15'
-  ga:
-    java:
-      lower: '1.0.0'
+  # TODO: Python must go to 1.0.20 or 1.1.0 when we GA, due to mistakenly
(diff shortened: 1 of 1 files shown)


gapic/packaging/api_defaults.yaml
```diff
@@ -27,13 +27,22 @@ generated_package_version:
   #       deleted but files still exist on the pypi servers and can prevent
   #       us from using the version number again.
   python:
-    lower: '0.15.4'
-    upper: '0.16dev'
+    default:
+      lower: '0.15.4'
+      upper: '0.16dev'
```
garrettjonesgoogle: One option to not be disruptive is to allow both old form (lower/upper) and new form (ga:lower/upper)

shinfan: Done

shinfan:
PTAL

garrettjonesgoogle:
LGTM, but don't submit until we have Artman / Toolkit support

shinfan:
@garrettjonesgoogle It turns out Java does not like a map that contains both version and another map at the same level. I have modified the structure so it is easier to work with in toolkit.

gapic/packaging/api_defaults.yaml
```diff
@@ -21,19 +21,24 @@ homepage: https://github.com/googleapis/googleapis
...
+    nodejs:
+      lower: '0.7.1'
+    php:
+      lower: '0.1.0'
+    ruby:
+      lower: '0.6.8'
+    java:
+      lower: '0.1.15'
+  ga:
```
garrettjonesgoogle: I'd prefer to push this up further if we can't do what I previously suggested: preserve the existing `generated_package_version` field, and add a new field named `generated_ga_package_version`. This prevents the Map of Map of Map structure that you see in Java.

shinfan: Done as you suggested. Updated toolkit and artman as well.

I considered using this approach (generated_ga_package_version) at first but I decided to go with the other way just to avoid adding another top-level field.

garrettjonesgoogle:
Ok the new form LGTM (of course, like previously, submission depends on approval of Artman and Toolkit)

PR #432, opened 2017-08-23
Add config for videointelligence v1beta2.

lukesneeringer:


lukesneeringer:
They are going to exist side by side.

PR #433, opened 2017-08-23
Add Ruby default release level

geigerj:


geigerj:
Context: I encountered an artman error regen'ing DLP due to this missing field. Not sure why this error hasn't come up before.

PR #434, opened 2017-08-24
Fix video intelligence target paths in Node.

lukesneeringer:
They were wrong previously. :-)


lukesneeringer:
Also, somehow the old `video_intelligence.yaml` was still in the tree. Corrected.

landrito:
I also forgot that I fixed a bug where hyphenated package names were causing an error in node.

While you're fixing videointelligence. can you change the package name [here](https://github.com/googleapis/googleapis/blob/master/google/cloud/videointelligence/v1beta2/videointelligence_gapic.yaml#L17) from `videointelligence.vx` to `video-intelligence.vx`?

PR #436, opened 2017-08-28
Fix broken link under service control page

cydu:
The link in this page is broken, found https://cloud.google.com/service-control/overview have the correct source of truth, so link it back to fix the broken link.


cydu:
I signed it!

chrisdunelm:
Old PR. Closing.

PR #438, opened 2017-08-30
spanner: Reducing total timeout of rpcs to equal rpc timeout.

vkedia:
@garrettjonesgoogle PTAL

### DIFF
--- a/google/spanner/v1/spanner_gapic.yaml
+++ b/google/spanner/v1/spanner_gapic.yaml
@@ -52,7 +52,7 @@ interfaces:
     initial_rpc_timeout_millis: 3600000
     rpc_timeout_multiplier: 1
     max_rpc_timeout_millis: 3600000
-    total_timeout_millis: 36000000
+    total_timeout_millis: 3600000
   methods:
   - name: CreateSession
     flattening:


google/spanner/v1/spanner_gapic.yaml
```diff
@@ -52,7 +52,7 @@ interfaces:
     initial_rpc_timeout_millis: 3600000
     rpc_timeout_multiplier: 1
     max_rpc_timeout_millis: 3600000
-    total_timeout_millis: 36000000
+    total_timeout_millis: 3600000
```
garrettjonesgoogle: In general, we have the total_timeout be a larger value than max_rpc_timeout to allow for at least one retry on a timeout. Setting them to the same value makes retries on timeouts impossible.

benwulfe: this is still a full one hour timeout for a single rpc call?
is this the timeout that will trigger if the tcp connection is dropped server side?  1 hr seems like an awfully big penalty for the client to pay.  thread based clients could have a significant portion of their threadpool blocked waiting for the timeout.

vkedia: @garrettjonesgoogle you are right that this will make retries on timeouts impossible. That is kind of what we prefer. We prefer that on the first rpc itself users should set timeout to whatever total time they can wait rather than setting a shorter timeout and then retrying the rpc. 
@benwulfe Most of the times, this timeout will trigger if the server did not finish the work in this much time. I think for cases like dropping tcp connection, grpc would fail immediately with unavailable.

benwulfe: @vkedia It depends.  I've recently seen cases on android where a dropped tcp connection between client and server did not register server side for 20 minutes due to the tcp timeout.
Read https://stackoverflow.com/questions/9449209/how-does-tcp-connection-terminate-if-one-of-the-machine-dies for more info.

vkedia: Ok I am not sure what gRPC does in these scenarios. From spanner point of view we do want really large deadlines for these rpcs since these can take arbitrary long. We have seen customer running into issues with the current values of these timeouts and unfortunately veneer does not provide a way for users to override these timeouts in all the languages.

garrettjonesgoogle: @vkedia I've been on the lookout for the following scenario and wanted to ask if it is the case for your API: do you need an auto-computed timeout that is relative to the data size?

vkedia: I don't think for any of our rpcs the timeout is dependent on the size of the request. It might be dependent on the size of the response for the streaming rpcs but you cannot no a priori how much data will be streamed back.

vkedia:
Any other concerns here or can this be merged?

PR #442, opened 2017-08-31
Speech 1.1 beta protos.

lukesneeringer:


gapic/api/artman_speech_v1_1beta1.yaml
```diff
@@ -0,0 +1,99 @@
+common:
+  api_name: speech
+  api_version: v1_1beta1
```
landrito: IIRC this should be v1p1beta1

lukesneeringer: Is this the thing that affects the endpoint? If so then yes.

landrito: For reference: https://cloud.google.com/apis/design/versioning

PR #443, opened 2017-08-31
Fix Ruby VideoIntelligence GAPIC code dir config

geigerj:
Previously was inconsistent with all other configs


geigerj:
Confirmed -- this only affects output directories

PR #447, opened 2017-09-05
Add logging-external-types package

eoogbe:


gapic/core/artman_logging_external_types.yaml
```diff
@@ -0,0 +1,38 @@
...
+  organization_name: google
+  proto_deps:
+    - google-common-protos
+    - google-iam-v1
+  import_proto_path:
+    - ${GOOGLEAPIS}
+  src_proto_path:
+    - ${GOOGLEAPIS}/google/appengine/legacy
+    - ${GOOGLEAPIS}/google/appengine/logging/v1
+    - ${GOOGLEAPIS}/google/appengine/v1
```
geigerj: We don't want all of the appengine protos, only the logging ones... I'm not sure whether the current syntax allows specifying a path to the proto file itself rather than to a directory. If not, you might need to modify `artman` to allow this.

eoogbe: Done

geigerj: A blacklist will be brittle solution since there's no way to update it automatically if a new proto ever gets pushed. Can we update artman to accept files as arguments to `src_proto_path` instead.

eoogbe: Done.

PR #453, opened 2017-09-11
NodeJS: Support protos being copied to gapic directory.

landrito:


landrito:
Closing. It seems we are still going to need this dep.

PR #457, opened 2017-09-15
Add artman-genfiles/ to .gitignore.

lukesneeringer:
The new `artman2` currently only runs from the root of this repo (I will fix this today). Ignore its output by default.


satishpatel2:
Network connections  problem

PR #463, opened 2017-09-15
bump java proto version to prepare proto/grpc release

neozwu:


bweston92:
Already gone above to `0.12` now so this can be closed?

PR #470, opened 2017-09-27
bump java proto version to 0.1.20 for new proto/grpc release

andreamlin:
New proto/grpc release in https://github.com/googleapis/api-client-staging/pull/416.


andreamlin:
PTAL.

PR #473, opened 2017-10-04
Update Python generator ID

pongad:
https://github.com/googleapis/toolkit/pull/1522 made the Python ID "py", not "python".

I believe the source of truth for this file is GitHub, not Google repo. So making the change here.


pongad:
Never mind, it's also sourced from google repo

PR #474, opened 2017-10-06
Run artman smoketests in googleapis circleci project

ethanbao:
Using the same smoketest script that artman uses. Different from the circleci task, we uses a stable artman image to run the smoketests. This way, we can be notified if one googleapis commit causes some artifact fail to generate. Currently, some artifacts fail to generate, causing noisyness. This task will be more useful once all artifacts in googleapis are working.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -13,3 +13,24 @@ jobs:
               git push private HEAD:master
             fi
     working_directory: /var/code/googleapis/
+  smoke-all:
+    docker:
+      - image: googleapis/artman:stable
+    steps:
+      - checkout
+      - run:
+          name: Run smoke tests
+          command: |
+            mkdir /tmp/reports
+            smoketest_artman.py --root-dir=/var/code/googleapis/ --log=/tmp/reports/smoketest.log
+      - store_test_results:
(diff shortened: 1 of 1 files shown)


geigerj:
Given that this repo is now entirely mechanically synchronized, wouldn't it be more useful to do this internally, upstream?

geigerj:
@ethanbao From offline discussion, I understand that there are difficulties in doing this upstream. However, I am concerned that alone this will not be beneficial unless we plan to set up an alert system. Team members don't do much (if any) manual monitoring of this repo.

ethanbao:
Setup a alert/notification is a different topic. Just because people doesn't follow the law doesn't mean there is no need to write the law. The law enforcement can be addressed separately. It would be nice if we can enforce something before bad things to happen. But sometimes it could be hard, so post-crisis enforcement will still benefit our stack. We can have monitoring as part of oncall rotation. At minimal, is we will be able to trace and check which PR/commit caused an artifact fail to build.

geigerj:
@ethanbao Acknowledged. I just think that at a pragmatic level this will not have as much impact as it should. We don't even have an effective process for triaging issues in this repo currently.

PR #479, opened 2017-10-26
Change Ruby package metadata defaults

geigerj:
- Bump googleauth dependency to 0.6.1 to pick up new credentials class
  required by GAPIC.
- Reset default version to 0.1.0. Brand-new package should have this
  version. All other packages should have their version maintained
  manually or through the artman config.


geigerj:
Whoops, wrong place.

PR #486, opened 2017-11-21
Firestore Field Names Documentation

chemelnucfin:
How do I render the documentation locally to see what it looks like?

@jba I corrected the regex expression, and changed the wording.  I think what I wrote is correct given what we discussed, but please feel free to correct me.  I hope that this is a bit clearer in describing what a field path is.


jba:
Thanks for doing this, but I think the source of truth for this is inside google. I believe an internal bug has already been filed.

chemelnucfin:
ok.

PR #491, opened 2018-01-16
getIamPolicy should be a GET and not a POST

bshaffer:


wora:
All files in this repo reflects Google's production services. So no breaking change is ever allowed. Changing a file in this repo has no effect on the running servers, and it would break all client libraries.

We could argue this was a design mistake, but that happened 4 years ago. At this point, we have to live with it.

PR #508, opened 2018-07-11
add default host and scopes

pongad:


wora:
I don't agree with the change. Google APIs are designed to be reusable by other people who want to implement these interfaces. We should not put any Google product specific information into the proto files. Such information should be store elsewhere.

All GCP APIs should support `cloud-platform` scope, so there is no need to put it into every file.

lukesneeringer:
This is not going onto master (note the base branch is `input-contract`).

That said, having metadata hardly precludes someone from reimplementing the interface, and we already have Google-specific information in protos, including public ones; the _names_ of the APIs (e.g. "Spanner") are Google trademarks, for example.

sebastianoconnell:
public class WebAppInterface { Kontekst mContext ; / ** Wywołaj interfejs i ustaw kontekst * / WebAppInterface ( Kontekst c ) {         mContext = c ; } / ** Pokaż toast ze strony internetowej * / @JavascriptInterface public void showToast ( String toast ) { Toast . makeText ( mContext , toast , Toast . LENGTH_SHORT ). show ();   
    

    
     

    

    
    
      
         
    } }

pongad:
I believe this is being done elsewhere.

PR #509, opened 2018-07-12
use latest artman image

alexander-fenster:
We don't use `stable` tag so let's use `latest`.


alexander-fenster:
Discussed offline with @garrettjonesgoogle so merging.

PR #510, opened 2018-07-25
Exclude APIs with (experimental) version 0 from CPP build

schlagert:
The main reason for this PR is that the recently committed Google Ads API does break the CPP client stub generation. Unfortunately, this is not caught by the CircleCI build/test script which only tests _Artman_-based generation.


schlagert:
I signed it!

chrisdunelm:
Old PR. Closing

PR #512, opened 2018-07-30
Add response field to HttRule

doroginin:
This PR adds optional field `response_body` from spec: https://cloud.google.com/service-management/reference/rpc/google.api#google.api.HttpRule

### DIFF
--- a/google/api/http.proto
+++ b/google/api/http.proto
@@ -297,6 +297,13 @@ message HttpRule {
   // present at the top-level of request message type.
   string body = 7;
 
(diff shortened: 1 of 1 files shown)


doroginin:
I think we need regenerate the correspondent
 http.pb.go too: https://github.com/google/go-genproto/blob/master/googleapis/api/annotations/http.pb.go

chrisdunelm:
PR no longer required. Closing.

doroginin:
> PR no longer required. Closing.

Why?

PR #516, opened 2018-08-21
Rollback remaining parts of #273

vam-google:
0) This clenup is necessary in the context of bazel-based gapic codegen work.
1) Partially the #273 PR was rolled back in #298, which made the remaining parts broken anyways (for example monitoring `BUILD` file reffers to rpc and annotations targets, which were removed in #298).
2) The `.gitmodules` point to 2 year old repository commits, boringssl link is broken.

Note, as part of bazel codegen integration this repository (googleapis) will be populated with corresponding `BUILD.bazel` files (NOT part of this commit), which may potentially break teams which import googelapis as new github repo (using `new_git_repository` workspace rule) and then overlay their custom build files on top of it. 

To fix that the teams which currently overlay custom build files on googleapis packages are expected to migrate to using the provided `BUILD.bazel` files instead (they should contain everything needed, like `proto_library`, `java_proto_library`, etc. definitions). In case if something is missing, it should be added directly to the `BUILD.bazel` files provided in googleapis, instead of overlaying custom build files. 

Practically, for teams which currently rely on custom build files, this means switching from `new_git_repository` to `git_repository` in their WORKSPACE file definitions, then fixing target naming differences (if any) and transferring missing stuff (if any) from custom `BUILD` files to googleapis' `BUILD.bazel` files.


vam-google:
@makdharma PTAL

PR #517, opened 2018-09-08
Input contract unstable

Aalmaho:


wora:
I don't think we can accept this PR. This repo contains API definition
matches the current Google services running in production. So every API
change must be done by each service owner and qualified in production
before the change is released here. So there should not be any PR goes
directly to this repo.

chrisdunelm:
Old PR. Closing.

PR #519, opened 2018-09-27
Add resource.path to Subscription message

andreamlin:
Just want a sanity check to verify that this is how ResourceTypes work.

Currently there's a circular reference between Subscription.name and Subscription. Nowhere, outside of comments, is there a definition for the resource path template that identifies a Subscription. This fixes that.

### DIFF
--- a/google/pubsub/v1/pubsub.proto
+++ b/google/pubsub/v1/pubsub.proto
@@ -607,7 +607,7 @@ message Subscription {
   // in length, and it must not start with `"goog"`.
   string name = 1 [
       (google.api.required) = true,
-      (google.api.resource).path = "projects/{project}/subscriptions/{sub}"];
+      (google.api.resource).path = "projects/*/subscriptions/*}"];
 
   // The name of the topic from which this subscription is receiving messages.
   // Format is `projects/{project}/topics/{topic}`.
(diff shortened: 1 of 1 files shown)


andreamlin:
PTAL

google/pubsub/v1/pubsub.proto
```diff
@@ -607,7 +607,7 @@ message Subscription {
   // in length, and it must not start with `"goog"`.
   string name = 1 [
       (google.api.required) = true,
-      (google.api.resource_type) = "google.pubsub.v1.Subscription"];
+      (google.api.resource).path = "projects/{project}/subscriptions/{sub}"];
```
chrisdunelm: I think this is needs to be: `projects/*/subscriptions/*`.
I agree it should be `resource` rather than `resource_type`.

wora:
The proto definitions in this repository are published by Google production
services. Any manual change will be overridden by next server change.
Please do not make any manual change. You can report an issue for the
service owner to fix bugs.

chrisdunelm:
This change is not merging into the `master` branch, but into the `input-contract` branch which is not managed by Google production services.
As I understand it, this should be good to merge after @lukesneeringer has looked at it.

PR #520, opened 2018-10-02
Fix cases where resource_type is incorrectly applied

chrisdunelm:


andreamlin:
I had the same questions!

PR #521, opened 2018-10-05
Add Pubsub annotations for resource sets and deleted_topic

andreamlin:
To make pubsub.proto match pubsub_gapic.yaml, we should make `Topic` a resource_set that contains 
- `topic` with path `projects/*/topics/*`
- `deleted_topic` with path `_deleted-book_`


andreamlin:
PTAL

PR #524, opened 2018-10-22
Cloud-Resource-Manager proto annotations

chrisdunelm:


google/cloud/resourcemanager/v2/folders.proto
```diff
@@ -201,11 +207,11 @@ message Folder {
 
   // Output only. The resource name of the Folder.
   // Its format is `folders/{folder_id}`, for example: "folders/1234".
-  string name = 1;
+  string name = 1 [(google.api.resource_definition).path = "folders/{folder}"];
```
andreamlin: So this assumes that each message type can have at most one `resource_definition`?

chrisdunelm:
This is now updated and may be mergable... PTAL

google/cloud/resourcemanager/v2/folders.proto
```diff
@@ -30,6 +30,12 @@ option java_package = "com.google.cloud.resourcemanager.v2";
 option csharp_namespace = "Google.Cloud.ResourceManager.V2";
 option php_namespace = "Google\\Cloud\\ResourceManager\\V2";
 
+// The "organization" resource is a common resource, currently defined in api-common-protos/.../annotations.proto
+// the "Folder" resource is defined within the "Folder" message in this file
+option (google.api.resource_set_definition) = {
+  name: "Parent",
+  resource_references: ["google.api.Organization", "Folder"]
```
andreamlin: `google.api.Organization` was removed in https://github.com/googleapis/api-common-protos/pull/20/files.

andreamlin: Is "Folder", not-fully-qualified, ok here because the resource definition on one of the Folder message' field doesn't have a `name` value specified so the default `name` is the message type name `Folder`?

If we had used the fully qualified name `google.cloud.resourcemanager.v2.Folder` reference instead, would that also be ok? Or should we only be referring to a Resource definition's `name` (even if it's derived from a message name)?

chrisdunelm: Thanks, yes, although I don't think `google.api.Organization` has a replacement yet; do you know if that's correct?

Whether a resource is defined using a top-level `resource_definition` or from within a proto message, it always gets the namespace of the package. So in this case it's fine to use `Folder` because in defaults to using the current namespace if not fully-qualified. Using the fully-qualified `google.cloud.resourcemanager.v2.Folder` would also be fine.

chrisdunelm:
Old PR. Closing.

PR #525, opened 2018-10-30
Update vision and pubsub proto annotations

chrisdunelm:
To match latest specification


andreamlin:
LGTM

Separate matter for future PR: can we pull api-common-protos:input-contract into this branch so that we can ensure that the protos compile?

PR #526, opened 2018-11-07
Remove references to google.api.Project; create resource def for projects/{project}

andreamlin:
`google.api.Project` was removed in https://github.com/googleapis/api-common-protos/pull/20/files.

### DIFF
--- a/google/pubsub/v1/pubsub.proto
+++ b/google/pubsub/v1/pubsub.proto
@@ -38,7 +38,7 @@ option (google.api.metadata) = {
 };
 
 option (google.api.resource_definition) = {
-  name: "project"
+  name: "Project"
   path: "projects/{project}"
 };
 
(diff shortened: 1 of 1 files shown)


andreamlin:
PTAL

andreamlin:
Project with a capital P.

chrisdunelm:
I don't think we want to do this.

In C# we require a common (global) definition of the `Project` and `Organization` resource types; partly so we don't get a breaking change when moving to the micro-generator; but more importantly because it is semantically correct - that is, all APIs using `Project` or `Organization` are referring to the same thing, so they should be the same type.

Exactly *where* these resource types should be defined is an open question.

andreamlin:
We can close this if https://github.com/googleapis/api-common-protos/pull/24 goes thru.

lukesneeringer:
Closed as out of date.

PR #527, opened 2018-11-07
DO NOT MERGE: Pubsub new annotations

landrito:


google/pubsub/v1/pubsub.proto
```diff
@@ -30,6 +30,15 @@ option java_outer_classname = "PubsubProto";
 option java_package = "com.google.pubsub.v1";
 option php_namespace = "Google\\Cloud\\PubSub\\V1";
 
+option (google.api.metadata) = {
+  product_name: "PubSub"
+  product_uri: "https://cloud.google.com/pubsub/"
+};
+
+option (google.api.resource_definition) = {
```
chrisdunelm: Project does need to be a common resource, not defined here.

landrito: Originally this resource_definition was going to be defined in `google/api/{something}.proto` so that if you imported the annotations the resource would implicitly be defined. IIUC this makes it such that generated class for the project resource will be common amongst those who use it.

landrito: Also, to be clear, I agree that this should be defined commonly elsewhere.

andreamlin: @lukesneeringer The Project message was removed in https://github.com/googleapis/api-common-protos/pull/20/files. Are there plans to reintroduce it, or are we not defining a common resource (there)?

google/pubsub/v1/pubsub.proto
```diff
@@ -417,7 +417,7 @@ message Topic {
   // must not start with `"goog"`.
   string name = 1 [
       (google.api.field_behavior) = REQUIRED,
-      (google.api.resource).path = "projects/{project_id}/topics/{topic_id}"];
+      (google.api.resource).path = "projects/{project}/topics/{topic}"];
```
landrito: Generally these should be `project_id` and `topic_id` but I think they need to stay project and topic to keep compatibility with the current gapic config.

chrisdunelm: As commented on in https://github.com/googleapis/api-common-protos/pull/24/files the C# generator currently auto-adds the `id` suffix if it's missing, so C# won't be broken by using `project_id` and `topic_id` here.
@andreamlin How does Java handle this?

CHANSAW:
wow

chrisdunelm:
Old PR. Closing.

PR #528, opened 2018-11-08
CMake implementation tested building for Linux, Mac, Windows, and Android

jwinarske:
The initial implementation can be found here:
  https://github.com/jwinarske/assistant-sdk-cpp/blob/cmake/cmake/build_googleapis.cmake


chrisdunelm:
Old PR. Closing.

PR #541, opened 2019-03-13
Add missing signature and field-behaviour to vision

chrisdunelm:


chrisdunelm:
@lukesneeringer I know this branch will soon be going away, but I'd like to keep it correct until then!

PR #543, opened 2019-03-27
Add bazel build instructions to README

eoogbe:


README.md
```diff
@@ -7,6 +7,32 @@ Google APIs and help you to utilize them more efficiently. You can also
...
+bazel build //...
+```
+
+To test all libraries:
+
+```
+bazel test //...
+```
+
+To build one library in one language:
```
vam-google: This is somewhat inaccurate. That target will crate a bazel-independent package (almost identical to what artman generates, the one with gradle scripts in it for java).  To build single lisbrary it would be something like:
```
bazel build //google/example/library/v1/...
```
That would build all target in the library (including the `google-cloud-library-v1-java` target).

I belive it is worth having both targets as separate items in the readme (i.e. so users know how to build everything for a single library and also know how to get artman-like pakcage for the single library explicitly).

eoogbe: Done

eoogbe: @vam-google Is package more accurate?

PR #544, opened 2019-04-01
Dropped unused load from BUILD file

mhutchinson:
The java_grpc_library was not used in this BUILD file. Removing this allows other repositories to reference the proto libraries in this package without importing the java grpc repo.


vam-google:
@mhutchinson Thanks for you (dependencies are evil, so removing redundant dependencies is always good)! The internal way is the right way. I'll close this one. Thanks.

PR #545, opened 2019-04-01
DO NOT MERGE: First pass logging config update

michaelbausor:
@andreamlin @lukesneeringer @chrisdunelm 

I haven't actually tested anything yet, so there are probably all sorts of syntax errors. I'll continue to update this PR, but thought I'd share now.

Promising sign is that we have a large net deletion of lines, despite logging being complex and needing to add a new .proto file for the resources.


chrisdunelm:
Experimental old PR. Closing.

PR #550, opened 2019-04-25
Add GAPIC Annotations for BigQuery V2

vam-google:


google/cloud/bigquery/v2/BUILD.bazel
```diff
@@ -0,0 +1,149 @@
+# This is an API workspace, having public visibility by default makes perfect sense.
```
andreamlin: Do we all have to include BUILD.bazel files?

andreamlin:
@vam-google Thank you for doing this!

Unfortunately we should wait to submit this because there's a hitch in how paging works for Bigquery, see https://github.com/googleapis/gapic-generator/issues/2738 for details. 

We should keep this open until we figure out how to resolve that issue, and then we can submit :)

PR #551, opened 2019-04-26
Translate API to gapic config v2

andreamlin:


google/cloud/translate/v3beta1/translate_gapic.yaml
```diff
@@ -110,126 +99,20 @@ interfaces:
...
-      name: glossary
-    timeout_millis: 60000
   - name: DeleteGlossary
-    flattening:
-      groups:
-      - parameters:
-        - name
-    required_fields:
-    - name
     retry_codes_name: idempotent
```
andreamlin: The default here would have been non_idempotent. The API might be wrong on this one, because it is server-side idempotent but probably not client-side idempotent.

chrisdunelm:
This looks good to me, although I haven't exhaustively checked every detail.

How identical is the generated code before/after this PR?

OmarAlhammadi1996:
> This does result in breaking changes for C# only.

PR #552, opened 2019-04-29
Tasks v2beta3 Proto Annotations + Gapic Config v2

hzyi-google:
Tried to generate gapics using bazel ` bazel test //google/cloud/tasks/v2beta3:tasks_go_gapic_test` and `bazel test //google/cloud/tasks/v2beta3:tasks_java_gapic_test`.
- Generation for Go was _successful_.
- Generation for Java _failed_. The error message seems to indicate resource name classes were not generated (`LocationName`, `QueueName`, `TaskName`).


hzyi-google:
PTAL

google/cloud/tasks/v2beta3/cloudtasks_gapic_v2.yaml
```diff
@@ -0,0 +1,129 @@
...
+    timeout_millis: 10000
+  - name: ListTasks
+    retry_params_name: default
+    timeout_millis: 10000
+  - name: GetTask
+    retry_params_name: default
+  - name: CreateTask
+    retry_params_name: default
+  - name: DeleteTask
+    retry_params_name: default
```
andreamlin: Can you add back in the `retry_codes_name: idempotent` line for DeleteTask? Otherwise it defaults to non-idempotent

google/cloud/tasks/v2beta3/cloudtasks.proto
```diff
@@ -357,7 +384,10 @@ message CreateQueueRequest {
   // The list of allowed locations can be obtained by calling Cloud
   // Tasks' implementation of
   // [ListLocations][google.cloud.location.Locations.ListLocations].
-  string parent = 1;
+  string parent = 1 [
```
andreamlin: `Queue queue = 2` below in this message should also be required, according to the old GAPIC yaml

google/cloud/tasks/v2beta3/cloudtasks.proto
```diff
@@ -511,7 +559,10 @@ message CreateTaskRequest {
   // `projects/PROJECT_ID/locations/LOCATION_ID/queues/QUEUE_ID`
   //
   // The queue must already exist.
-  string parent = 1;
+  string parent = 1 [
```
andreamlin: I think `CreateTaskRequest.task` should also be required, according to the GAPIC config.

google/cloud/tasks/v2beta3/cloudtasks_gapic_v2.yaml
hzyi-google: Also added back some more `retry_codes_name: idempotent` that were deleted by accident, and removed one `retry_codes_name: non-idempotent`.

hzyi-google:
PTAL

andreamlin:
@hzyi-google Thanks a ton for doing this!

PR #567, opened 2019-07-25
Highlight need for a summary comment for SpeakerDiarizationConfig.

tseaver:
Lack of it breaks downstream code generation in Python -- see https://github.com/googleapis/google-cloud-python/pull/8780.


tseaver:
Obviously I don't expect this change to land as typed.  Please close the PR once the upstream change propagates here.

PR #571, opened 2019-08-25
Fix typo in publish_build_event.proto

endobson:


chrisdunelm:
Old PR. And changes here will be overwritten by an auto-copier. Closing.

PR #572, opened 2019-08-25
Remove old comment in publish_build_event.proto

endobson:
This line refers to fields which don't exist.


chrisdunelm:
Old PR. Closing.

PR #573, opened 2019-08-27
Remove unused import in watch.proto

endobson:


chrisdunelm:
This file is auto-synced from a Google-internal repo, so this edit will be overwritten. Closing.

PR #575, opened 2019-09-10
speedup CI by saving one .tgz file

alexander-fenster:
With the current size of `googleapis`, CI takes about an hour to run, and sometimes 2 hours or more to save all the artifacts (generated files for all languages × all APIs). Let's try to speed up the second part by saving just one big `.tar.gz` file instead.

The log will still be available in CircleCI artifacts page since `/tmp/reports` folder is saved separately.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -43,4 +43,3 @@ workflows:
             branches:
               only:
                 - master
-                - speedup-ci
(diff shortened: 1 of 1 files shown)


alexander-fenster:
Before:

image

After:

image

Saved an hour (sometimes more) per task :)

PR #615, opened 2020-09-02
How to download drive files through HTTP requests with the access_token included in the url?

zyadomer999:
Can someone please help me on finding a way to send an HTTP request to Google Drive to download a file with the access token included in the url?

I searched a lot in the internet and found only this way to send an access_token in URL: "https://www.googleapis.com/drive/v3/files/[fileID]?access_token=ya29.a0AfH6SMA4FU********&key=[apiKey]"

As you know, it only returns the metadata of the file because it does not contain alt=media. If you add "alt=meida" to the url, it fails to load and returns nothing.

It can be easily fixed with passing the access_token as a header. However, I cannot do that because I am using the url as a "src" of HTML and as a source for video segments in .m3u8 file.

Any suggestion?


chrisdunelm:
@zyadomer999 Please re-ask this question as an issue, not a PR.
Closing.

PR #616, opened 2020-09-02
Move C# to using the micro-generator

chrisdunelm:


chrisdunelm:
@vam-google Please can you take a look at this draft PR, and confirm whether this is what you expect for the C# micro-generator change?
I've also included the bazel change required for just the Language API, to show what will be required.

WORKSPACE
```diff
@@ -316,6 +316,14 @@ php_gapic_repositories()
 # C#
 ##############################################################################
 
+# Required to access the C#-specific common resources config.
+http_archive(
+    name = "google_cloud_dotnet",
+    urls = ["https://github.com/googleapis/google-cloud-dotnet/archive/1b0afcc7b11b02aaea3456f822c8030d5e9f361c.zip"],
```
vam-google, line 322: Is it what Ads is using? I mean it will work, but it basically creates a circular dependency between the repositories. This repo is used to create `google-cloud-dotnet`, and it imports `google-cloud-dotnet` as a dependency at the same time. Does this whole import exists to import a single file (CommonResourcesConfig.json)? If yes, why is it located in google-cloud-dotnet repostory? (not in googleapis, or gapic-generator-csharp for example)?

google/cloud/language/v1/BUILD.bazel
```diff
@@ -341,10 +341,9 @@ csharp_grpc_library(
 
 csharp_gapic_library(
     name = "language_csharp_gapic",
-    src = ":language_proto_with_info",
-    gapic_yaml = "language_gapic.yaml",
-    package = "google.cloud.language.v1",
-    service_yaml = "//google/cloud/language:language_v1.yaml",
+    srcs = [":language_proto_with_info"],
+    grpc_service_config = "language_grpc_service_config.json",
+    common_resources_config = "@google_cloud_dotnet//:CommonResourcesConfig.json",
```
vam-google, line 346: This looks like something which remains constant for all `charp_gapic_library` targets. If this is the case and if `CommonResourcesConfig` resided in `gapic-generator-csharp` (the same repository which defines csharp_gapic_library now), then it could be included by default every time in the `csharp_gapic_library` implementation. This would let us making the surface of the rule simpler (minus one argument) but most importantly remove the dependency on `google-cloud-dotnet` from `googleapis`.

vam-google (commented):
Overall looks good, but I'm a little bit concerned by the introduced dependency on `google-cloud-dotnet` repo.
It seems like it (dependency) exists solely for one file (`CommonResourcesConfig.json`), but most importantly it introduces a conceptual circular dependency between the repositories (check the corresponding review comment for details).

chrisdunelm:
Thanks for the review.
Yes, the `google_cloud_dotnet` import exists just to access the one file. The file could potentially be moved (or copied) into the generator repo.
That file (CommonResourcesConfig) will generally be common for all cloud APIs, but it is required as a parameter for the generator as extra customization will be required for some libraries - especially non-cloud and third-party APIs.
I'll update this PR once I've considered how to better access the the common resources file.

chrisdunelm:
@vam-google Updated to now refer to the common resources config in the dotnet-gax repo, after discussions with Jon.

vam-google (approved):
LGTM

chrisdunelm:
Closing this PR as it needs changing internally to google which will then be mirrored here.

PR #620, opened 2020-09-22
feat: Add Java microgen rules to imports

miraleung:
This will enable local development and building out the integration test framework on actual protos.


WORKSPACE
```diff
@@ -141,8 +141,17 @@ apple_support_dependencies()
 ##############################################################################
 # Java
 ##############################################################################
+# Java microgenerator.
+_java_microgenerator_version = "3f098c4e21daa7b211cd43e6c4af03d627e754e4"  # Will change often in dev.
```
vam-google, line 145: Please don't use variables in workspace file. This lets to save the one duplicate occurrence of this git hash, but it  converts the workspace file from being declarative to being imperative. It is not a big deal here, but it opens a door for treating this WORKSPACE file as an "program", complicating this already very hard to parse file.

WORKSPACE
```diff
@@ -141,8 +141,17 @@ apple_support_dependencies()
...
+
+http_archive(
+    name = "com_google_api_generator",
```
vam-google, line 148: I see you are following the same naming convention used in the monolith generator (the name of the bazel import corresponds to the name of the root java package of the generator). The other microgenerators simply use `gapic_generator_<language>` naming scheme, which has proven itself being less confusing to the users of this repo. Please consider using the same naming as the other microgenerators.

repository_rules.bzl
```diff
@@ -151,6 +151,24 @@ def switched_rules_by_language(
         "@com_google_api_codegen//rules_gapic/java:java_gapic_pkg.bzl",
     )
 
+    # Java microgenerator.
+    # These will eventually replace the monolith Java rules.
+    rules["java_microgenerator_gapic_library"] = _switch(
```
vam-google, line 156: We have been using the `2` suffix to distinguish between microgenerator and monolith while both are present in the imports (see [python_gapic_library2](https://github.com/googleapis/googleapis/blob/master/repository_rules.bzl#L173)) The idea was that microgenerator is an implementation detail, which is not important for the users of this repository. The only thing which matters is that it is some sort of a new version of the same rule, that is why it is called as `java_gapic_library2`. 

Please consider using the same naming convention here.

vam-google (approved):
LGTM, with a few minor comments.

kaitou66 (approved):

miraleung:
This will work when [this PR](https://github.com/googleapis/gapic-generator-java/pull/340) is merged in and the hash here is updated.

jskeet:
Please note that the set of 6 commits here caused some confusing downstream autosynth changes. It would be best to rebase from the default branch and force push onto your branch before then rebasing back onto the default branch, so we only end up with the relevant commits in the history.

PR #623, opened 2020-10-15
docs: change relative URLs to absolute URLs to fix broken links

laljikanjareeya:
Fixes https://github.com/googleapis/nodejs-redis/issues/385

### DIFF
--- a/google/cloud/managedidentities/v1/resource.proto
+++ b/google/cloud/managedidentities/v1/resource.proto
@@ -71,7 +71,7 @@ message Domain {
   map<string, string> labels = 2 [(google.api.field_behavior) = OPTIONAL];
 
   // Optional. The full names of the Google Compute Engine
-  // [networks](/compute/docs/networks-and-firewalls#networks) the domain
+  // [networks](https://cloud.google.com/compute/docs/networks-and-firewalls#networks) the domain
   // instance is connected to. Networks can be added using UpdateDomain.
   // The domain is only available on networks listed in `authorized_networks`.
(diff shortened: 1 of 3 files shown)


google/cloud/managedidentities/v1/resource.proto
```diff
@@ -71,7 +71,7 @@ message Domain {
   map<string, string> labels = 2 [(google.api.field_behavior) = OPTIONAL];
 
   // Optional. The full names of the Google Compute Engine
-  // [networks](/compute/docs/networks-and-firewalls#networks) the domain
+  // [networks](https://cloud.google.com/compute/docs/networks-and-firewalls#networks) the domain
```
frankyn, line 74: @danoscarmike do you know if we would need to make this change in the Redis protos instead?

danoscarmike: Yep, we should do this upstream. This change will get blown away by the next sync.

JustinBeckwith: @frankyn just checking in - anyone cut a CL here yet?

frankyn: No one has cut a CL yet. Apologies, I missed these two last responses.

frankyn: Looks like it got resolved in: https://github.com/googleapis/googleapis/commit/8be3770465a6cf19bea97a84b3e82b113b0239b6

PR #624, opened 2020-10-19
Add schema for providing GAPIC metadata

vchudnov-g:
This allows each (API,version,language) combination to provide a
mapping between RPCs and library methods invoking those RPCs.


gapic/metadata/gapic_metadata.proto
```diff
@@ -0,0 +1,84 @@
...
+  // for the transport and service specified in the containing
+  // structures.
+  message MethodList {
+    // List of methods for a specific proto-service client in the
+    // GAPIC. These names should be formatted as they appear in the
+    // source code.
+    repeated string methods = 1;
+  }
+
+}
```
alexander-fenster, line 84: could you please add a new line?

gapic/metadata/gapic_metadata.proto
```diff
@@ -0,0 +1,84 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+syntax = "proto3";
+
+package google.gapic.metadata;
```
alexander-fenster, line 18: if it's in `package google.gapic.metadata`, should it go to `google/gapic/` instead of just `gapic/`?

vchudnov-g: I was wondering. That would be nice in order to parallel the directory structure.

OTOH, is it really necessary? We have the `gapic/` directory here, and it would be nice to not get people confused into thinking "gapic" is a Google service....

gapic/metadata/gapic_metadata.proto
vchudnov-g: Done

gapic/metadata/gapic_metadata.proto
alexander-fenster: I don't insist here - we can as well leave it here.

alexander-fenster (approved):

gapic/metadata/gapic_metadata.proto
```diff
@@ -0,0 +1,84 @@
...
+  // Schema version of this proto. Current value: 1.0
+  string schema = 1;
+
+  // Any human-readable comments to be included in this file.
+  string comment = 2;
+
+  // Computer language of this generated language. This must be
+  // spelled out as it spoken in English, with no capitalization or
+  // separators (e.g. "csharp", "nodejs").
+  string language = 3;
```
martinvarelaj, line 32: What about making language an enum instead of a free string?

PR #631, opened 2020-12-08
rebase spanner_cmek

mr-salty:
Bring this branch up to date with `master`

This is needed to rebase the C++ client due to intervening proto changes (specifically `UpdateDatabaseDdlMetadata::throttled`)


mr-salty:
ugh, maybe it's easier to do this just by pushing directly to the branch instead of using a PR?

It should be easy - there were no conflicts and I tested it against the C++ client.

mr-salty:
I'm going to close this, we're using the cmek + pitr branch anyways... I'll leave it up to @skuruppu if we need to rebase this.

BTW, would it be possible to just have one branch for all unreleased protos?  We now have at least 7 which is pretty unmanageable, especially if we want to do combinations of features (like, tagging and rpc priority used to be in one branch but now they're separate, so the clients need to make 2 separate branches as well unless we make a combined branch like CMEK + PITR)

skuruppu:
Just to confirm, we'll try this out from next year onwards to keep all features in one branch. Let's see how that goes.

PR #635, opened 2020-12-30
feat(spanner): add RequestOptions support

themaximumspension:
Needed


fufedy (approved):
Come over here and chat

PR #636, opened 2021-01-05
chore: autoclose PRs in [googleapis]

miraleung:


.github/mistaken-pull-closer.yml
```diff
@@ -0,0 +1,16 @@
...
+# by probot.
+# Default behavior: Close all PRs.
+filters:
+  - true
+
+# The message to post to the closed PR.
+commentBody: |
+  Thanks for your contribution!  Unfortunately, we don't use GitHub pull
+  requests to manage code contributions to this repository.  Instead, please
+  see [CONTRIBUTING.md](../blob/master/CONTRIBUTING.md) which provides full
```
chingor13, line 12: This relative path might not be correct relative to the pull request page. It might be best to make this the full url
```suggestion
  see [CONTRIBUTING.md](https://github.com/googleapis/googleapis/blob/master/CONTRIBUTING.md) which provides full
```

chingor13: Also, the CONTRIBUTING.md doesn't really say what to do instead of opening a pull request here.

.github/mistaken-pull-closer.yml
```diff
@@ -0,0 +1,16 @@
+# The JSONPath filter expression used to identify which PRs to close.
+# The data filtered is the pull request data along with other metadata passed in
+# by probot.
+# Default behavior: Close all PRs.
+filters:
+  - true
```
chingor13, line 6: If we merge this, we won't be able to modify this file again (the pull request would be closed :rofl:). Perhaps we want to add filters?

chingor13 (changes requested):

SurferJeffAtGoogle (commented):
Why is this a pull request instead of a change list to google3/third-party/googleapis?

miraleung:
> Why is this a pull request instead of a change list to google3/third-party/googleapis?

Because my earlier CL to third_party got ignored by Copybara. It seems that all dot-named directories are ignored as a rule.

.github/mistaken-pull-closer.yml
miraleung: Should we say this repo doesn't accept contributions? Any other suggestions?

SurferJeffAtGoogle:
> > Why is this a pull request instead of a change list to google3/third-party/googleapis?
> 
> Because my earlier CL to third_party got ignored by Copybara. It seems that all dot-named directories are ignored as a rule.

Oh.  I guess that's a feature of copybara.  Thanks!

.github/mistaken-pull-closer.yml
alexander-fenster: In this repo we can always make a direct commit, they are not forbidden.

tydazzz1 (approved):

.github/mistaken-pull-closer.yml
miraleung: After digging, it doesn't look like there's a way to filter by usergroup or a similar sensible attribute.

dwsupplee:
@miraleung this seems to be taken care of now with the updates to the copybara job we landed recently.

Malik221184 (approved):

Malik221184 (approved):

mont1985:
@``

jskeet:
Closing.

PR #644, opened 2021-03-22
chore: add php rules to longrunning

bshaffer:
In order to update PHP's longrunning GAPIC clients, add BAZEL rules


bshaffer:
This PR breaks backwards compatibility due to the PHP GAPIC Client's handling of moving "required" parameters to "optional". Closing as this will require a more robust fix.

PR #658, opened 2021-06-06
Output file from protoc should be set to service.descriptor

nmandya:
The README.md file's instructions for running the `protoc` command specify the output service descriptor file name as `service.descriptors`. Since the extension of the filename is neither one of `.pb`, `.descriptor` or `.proto.bin`, the file gets read as a text file instead of a binary file. And because of that, the `gcloud endpoins services deploy` command fails with the following error:

```
'utf-8' codec can't decode byte 0xd7 in position 1: invalid continuation byte
```

This pull request is to change the instructions in the `README.md` file to generate the service descriptor file with the extension as `.descriptor`.

### DIFF
--- a/google/example/endpointsapis/README.md
+++ b/google/example/endpointsapis/README.md
@@ -89,11 +89,11 @@ Run the following commands to generate proto descriptor and push the
 descriptor and service configuration to the Service Management API.
 
 ```shell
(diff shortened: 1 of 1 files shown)


nmandya:
@googlebot I signed it!

PR #676, opened 2021-10-21
Remove unused proto google/protobuf/any.proto from cloud_speech.proto

dweekly:
Protoc throws a warning when called against this proto:
google/cloud/speech/v1p1beta1/cloud_speech.proto:24:1: warning: Import google/protobuf/any.proto is unused.

No clear usage inside the proto, so suggest eliding.


jskeet:
Thanks for your contribution!  Unfortunately, we don't use GitHub pull
requests to manage code contributions to this repository.  The code in
this repository is pushed from internal source control in Google.

dweekly:
No problem; can I request then that published protos like cloud speech only include imports that are used in the exported version? Otherwise downstream tooling will produce warnings.

PR #677, opened 2021-10-21
Remove Unused Proto Imports from cloud/speech/v1p1beta1/resource.proto

dweekly:
Note that protoc complains when processing this proto file:

google/cloud/speech/v1p1beta1/resource.proto:21:1: warning: Import google/api/annotations.proto is unused.
google/cloud/speech/v1p1beta1/resource.proto:20:1: warning: Import google/protobuf/timestamp.proto is unused.

These don't appear to be used here, so suggest eliding.

### DIFF
--- a/google/cloud/speech/v1p1beta1/resource.proto
+++ b/google/cloud/speech/v1p1beta1/resource.proto
@@ -17,8 +17,6 @@ syntax = "proto3";
 package google.cloud.speech.v1p1beta1;
 
 import "google/api/resource.proto";
-import "google/protobuf/timestamp.proto";
-import "google/api/annotations.proto";
 
 option cc_enable_arenas = true;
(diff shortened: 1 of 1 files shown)


jskeet:
Thanks for your contribution!  Unfortunately, we don't use GitHub pull
requests to manage code contributions to this repository.  Instead, please
see [CONTRIBUTING.md](../blob/master/CONTRIBUTING.md) which provides full
instructions on how to get involved.

dweekly:
No problem; can I request then that published protos like cloud speech only include imports that are used in the exported version? Otherwise downstream tooling will produce warnings.

jskeet:
I'm afraid that's rather more complex than it sounds, for reasons I can't easily go into. While I normally don't like saying "just ignore the warnings", I believe it's the right course of action in this particular case.

dweekly:
OK. I get that things are sometimes pretty complex internally
(epitaphs/dweekly).

Happy Monday!

PR #680, opened 2021-11-10
chore: Add C# generation for Compute v1

jskeet:
I expect this change will eventually need to be in internal source control, but I'm hoping that creating a PR is the right way forward for now.


jskeet:
Reopened as this is a real PR :)

vchudnov-g (approved):

PR #684, opened 2021-11-18
feat: Switch to string enums for compute

vam-google:
Also update grpc_service configs and introduce gapic_yaml for java clients to override default LRO configuration (which has too way too long timeouts).


WORKSPACE
```diff
@@ -381,10 +381,12 @@ gapic_generator_ruby_repositories()
 # Discovery
 ##############################################################################
 
+_disco_to_proto3_converter_version = "91953d241824df417b6d4e7c2292e4622b4af432"
```
alexander-fenster, line 384: Can we wait for this change to come from the sync?

api-index-v1.json
```diff
@@ -30782,7 +30782,7 @@
       "options": {
         "csharp_namespace": {
           "valueCounts": {
-            "Google.Cloud.Compute.V1Small": 1
+            "Google.Cloud.Compute.V1small": 1
```
alexander-fenster, line 30785: I would expect this one will be autogenerated?

alexander-fenster (approved):

api-index-v1.json
vam-google: hm... I don't even know how this got here, I did not touch this file... good catch.###

WORKSPACE
vam-google: It does not matter which one comes first, but yes, there is a pending internal cl with this change (you are added as a reviewer)

PR #685, opened 2021-11-19
feat: Switch to string enums for compute

vam-google:
Also introduce gapic_yaml for java clients to override default LRO configuration (which has too way too long timeouts).

Also cleanup regenerate other files (grpc_service_config and service yaml)


vchudnov-g (approved):
LGTM, but maybe wait for @noahdietz as well.

noahdietz (approved):

PR #686, opened 2021-11-24
fix: make parent_id fields required compute move and insert methods

vam-google:
Note that the order in `parameterOrder` property in Json file matters and must remain consistent. If we want to flip the order `firewallPolicy` and `parentId` this is the last chance to do so.

### DIFF
--- a/google/cloud/compute/v1/compute.proto
+++ b/google/cloud/compute/v1/compute.proto
@@ -8983,7 +8983,10 @@ message InsertFirewallPolicyRequest {
   FirewallPolicy firewall_policy_resource = 495049532 [(google.api.field_behavior) = REQUIRED];
 
   // Parent ID for this request. The ID can be either be "folders/[FOLDER_ID]" if the parent is a folder or "organizations/[ORGANIZATION_ID]" if the parent is an organization.
-  optional string parent_id = 459714768 [(google.cloud.operation_request_field) = "parent_id"];
+  string parent_id = 459714768 [
+    (google.api.field_behavior) = REQUIRED,
+    (google.cloud.operation_request_field) = "parent_id"
+  ];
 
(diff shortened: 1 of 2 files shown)


google/cloud/compute/v1/compute.proto
```diff
@@ -26645,7 +26651,7 @@ service FirewallPolicies {
     option (google.api.http) = {
       post: "/compute/v1/locations/global/firewallPolicies/{firewall_policy}/move"
     };
-    option (google.api.method_signature) = "firewall_policy";
+    option (google.api.method_signature) = "firewall_policy,parent_id";
```
vchudnov-g, line 26654: Do we want this signature to be consistent with the previous one, so that aprent_id is always at the front (or always at the back)?

google/cloud/compute/v1/compute.v1.json
```diff
@@ -16891,7 +16891,8 @@
         },
         "move": {
           "parameterOrder": [
-            "firewallPolicy"
+            "firewallPolicy",
```
vchudnov-g, line 16894: We should ping GCE as to the desired parameter order. What do other RPCs do with parentId?

vchudnov-g (approved):
A couple of questions, but otherwise LGTM. Thanks for doing this!

google/cloud/compute/v1/compute.v1.json
vchudnov-g: it seems like we only have parentID for firewallpolicies and global operations. It seems to me it would make sense to have parentId be the first parameter, if it is required?

google/cloud/compute/v1/compute.proto
vam-google: It is already consistent. For the other one the trailing parameter is body parameter. Body parameter is always the last one. This method des not have body, that is why it does not have a parameter after parent_id.

vchudnov-g: ok, cool

PR #690, opened 2021-12-02
chore(compute): include Go GAPIC targets

noahdietz:


google/cloud/compute/v1/BUILD.bazel
```diff
@@ -217,6 +217,55 @@ java_gapic_assembly_gradle_pkg(
...
+    ],
+)
+
+go_gapic_library(
+    name = "compute_go_gapic",
+    srcs = [":compute_proto_with_info"],
+    grpc_service_config = "compute_grpc_service_config.json",
+    importpath = "cloud.google.com/go/compute/apiv1;compute",
+    service_yaml = "compute_v1.yaml",
+    transport = ["rest"],
```
vam-google, line 246: Is it already supported by the rules? If yes, then great, I just missed the moment when it was added.

google/cloud/compute/v1/BUILD.bazel
```diff
@@ -217,6 +217,55 @@ java_gapic_assembly_gradle_pkg(
...
+
+go_test(
+    name = "compute_go_gapic_test",
+    srcs = [":compute_go_gapic_srcjar_test"],
+    embed = [":compute_go_gapic"],
+    importpath = "cloud.google.com/go/compute/apiv1",
+)
+
+# Open Source Packages
+go_gapic_assembly_pkg(
```
vam-google, line 260: Just as a sanity check - we do not need a transport parameter for assembly, do we (mayb e neded to generated proper build files and stuff)?

vam-google (approved):
LGTM, just a few questions.

google/cloud/compute/v1/BUILD.bazel
noahdietz: Yep, it has been supported for some time, it is just omitted from build_gen.

google/cloud/compute/v1/BUILD.bazel
noahdietz: Nope we do not need transport parameter for assembly. The `transport` param just toggles which transport interface implementations to generate. REST and/or gRPC. The code all goes into the same files, etc.

jirayupiwon8:
example.v1.Messaging.GetMessage

Dortega2250 (commented):
Ayuda

a2425rdl (approved):

PR #691, opened 2021-12-03
fix!: handle GCE enum name start with uppercase IPProtocol

summer-ji-eng:
Note: This is a *BREAKING CHANGE*.

The change is auto generated by latest `disco-to-proto3-converter`. The fix PR: https://github.com/googleapis/disco-to-proto3-converter/pull/62.

Enum type IPProtocol name(start with uppercase not lowercase) cause the protobufjs compiling error with type name duplicate.

To fix it, we need to rename enum name `IPProtocol` to `IPProtocolEnum`. This enum name should not affect other languages. Tested on googleapis/googleapis See googleapis/googleapis#689

Command to update generated compute.proto:
```
$ bazel build //google/cloud/compute/v1:compute_gen
$ cp bazel-bin/google/cloud/compute/v1/compute_gen.proto google/cloud/compute/v1/compute.proto
```

### DIFF
--- a/google/cloud/compute/v1/compute.proto
+++ b/google/cloud/compute/v1/compute.proto
@@ -615,7 +615,7 @@ message Address {
     // VM internal/alias IP, Internal LB service IP, etc.
     GCE_ENDPOINT = 230515243;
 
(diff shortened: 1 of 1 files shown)


vchudnov-g (approved):
nit: PR title: "GCP" → "GCE"

PR #701, opened 2022-02-11
fix: fix breaking changes in Compute API

vam-google:


unicrone72:
Thanks so much , this is greatly appreciated  ..

Sent from Mail<https://go.microsoft.com/fwlink/?LinkId=550986> for Windows

PR #705, opened 2022-02-25
chore: disable Python snippetgen for compute

parthea:


parthea:
https://github.com/googleapis/python-compute/pull/218#issuecomment-1044559224

vam-google (approved):
LGTM

PR #712, opened 2022-04-12
fix(compute): remove proto3_optional from parent_id

noahdietz:
When I made this field REQUIRED again, I forgot to remove the proto3_optional label as well.


vchudnov-g (approved):
Oops, sorry I missed that in my review.

PR #714, opened 2022-04-20
fix(compute): revert proto3_optional, required removal on parent_id

noahdietz:
Fixes accidental breaking change in https://github.com/googleapis/googleapis/commit/249e9a119d24a5cbda66e4f8a2063bf515dae9c0#r71665196.


vam-google (approved):
LGTM, but cat you please remove exclamation mark from conventional commit message. It will make automation start eleasing compute libraries 2.x which I don't think what we want to do.

noahdietz:
> but cat you please remove exclamation mark from conventional commit message

Done!

Ernest06 (commented):
Delete ma accounts and all

PR #721, opened 2022-05-19
chore: Add DIREGAPIC Updater github action

vam-google:


.github/workflows/diregapic.yaml
```diff
@@ -0,0 +1,50 @@
...
+      run: |
+        bazel build //google/cloud/compute/v1:compute_gen
+        cp bazel-bin/google/cloud/compute/v1/compute_gen.proto google/cloud/compute/v1/compute.proto
+        bazel build //google/cloud/compute/v1:compute_grpc_service_config_gen
+        cp bazel-bin/google/cloud/compute/v1/compute_grpc_service_config_gen.json google/cloud/compute/v1/compute_grpc_service_config.json
+        bazel build //google/cloud/compute/v1:compute_gapic_gen
+        cp bazel-bin/google/cloud/compute/v1/compute_gapic_gen.yaml google/cloud/compute/v1/compute_gapic.yaml
+        echo api_changes=$(git diff-index --shortstat HEAD) >> $GITHUB_ENV
+    - name: Build GAPIC clients
+      if: contains(env.api_changes, 'files')
```
alexander-fenster, line 31: It might say `1 file changed` (not `files`).

.github/workflows/diregapic.yaml
```diff
@@ -0,0 +1,50 @@
...
+      run: |        
+        bazel build //google/cloud/compute/v1/...
+        bazel build //google/cloud/compute/v1/...
+    - name: Create PR
+      uses: googleapis/code-suggester@v2
+      env:
+        ACCESS_TOKEN: ${{ secrets.YOSHI_CODE_BOT_TOKEN }}
+      with:
+        command: pr
+        upstream_owner: vam-google
```
alexander-fenster, line 41: This is not good, maybe we coud use any existing bot account?

.github/workflows/diregapic.yaml
```diff
@@ -0,0 +1,50 @@
...
+        upstream_repo: googleapis
+        title: 'chore: [DIREGAPIC] Update API definitions'
```
alexander-fenster, line 43: This is not good because the bots won't consider the `chore` commit as something that requires release.  Until we are able to generate good release messages (it's under consideration), let's use `feat:` instead to trigger releases.

alexander-fenster (approved):

.github/workflows/diregapic.yaml
vam-google: good catch

.github/workflows/diregapic.yaml
vam-google: Good catch. This is I just forgot to change it from my testing. It shoudl be googleapis.

.github/workflows/diregapic.yaml
vam-google: Ok, changed to `feat`

PR #734, opened 2022-09-07
Clarify the default for task_count_per_node

Shnatsel:
As per the discussion with @soj-hub, Batch actually defaults to 1 task per VM instead of using some heurtistic for packing multiple tasks on a VM. This clarifies the comment on the proto.

### DIFF
--- a/google/cloud/batch/v1/job.proto
+++ b/google/cloud/batch/v1/job.proto
@@ -430,8 +430,7 @@ message TaskGroup {
   repeated Environment task_environments = 9;
 
   // Max number of tasks that can be run on a VM at the same time.
-  // If not specified, the system will decide a value based on available
-  // compute resources on a VM and task requirements.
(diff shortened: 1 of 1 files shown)


google/cloud/batch/v1/job.proto
```diff
@@ -430,8 +430,7 @@ message TaskGroup {
   repeated Environment task_environments = 9;
 
   // Max number of tasks that can be run on a VM at the same time.
-  // If not specified, the system will decide a value based on available
-  // compute resources on a VM and task requirements.
+  // Defaults to 1 if not specified.
```
victoriaquasar, line 433: Is this the current behavior? Could you link to the part where this happens?

jskeet:
Closing as changes to protos mustn't be authored here: this repo is effectively a one-way sync from Google internal systems.

I've contacted @soj-hub internally.

PR #742, opened 2022-09-24
chore: regenerate API index

sr21780:


sr21780:
Someone has hacked all of my accounts and it's transferring to each device so I'm just trying to fix this

PR #756, opened 2022-11-18
Fix typo in publish_build_event.proto

SinOverCos:
publishd -> published


parthea:
Thanks for reporting this issue @SinOverCos, this will be fixed at the source in cl/489683023

PR #766, opened 2023-01-13
chore: Update Compute BUILD file to specify service config for C#

jskeet:
This is what's already specified when generating locally, so should be a no-op.


viacheslav-rostovtsev (approved):
LGTM

PR #777, opened 2023-02-08
fix: adds service yaml to v1p1beta1.

atulep:
Adds missing service YAML for speech API.


vchudnov-g (approved):

atulep:
The fix is correct, but this is an inappropriate place for it. We followed up internally.

PR #782, opened 2023-02-23
Fix Cloud Vision Go package path

wyattanderson:
I know this PR probably won't be accepted directly, but I can't figure out a better way to surface this issue.

The current Go package path for the `google.cloud.vision.v1` Protobuf definitions is incorrect. `cloud.google.com/go/vision/apiv1/visionpb` does not exist as a Go module. When published to `googleapis/google-cloud-go`, a `v2/` path segment is added (I believe this occurred in this pull request: https://github.com/googleapis/google-cloud-go/pull/6072). This complicates depending on this Protobuf definition because the generated Go code will be incorrect unless the Go package dependency is overridden out-of-band.

Maybe @quartzmo can look at this as the original author.

### DIFF
--- a/google/cloud/vision/v1/geometry.proto
+++ b/google/cloud/vision/v1/geometry.proto
@@ -17,7 +17,7 @@ syntax = "proto3";
 package google.cloud.vision.v1;
 
 option cc_enable_arenas = true;
-option go_package = "cloud.google.com/go/vision/apiv1/visionpb;visionpb";
+option go_package = "cloud.google.com/go/vision/v2/apiv1/visionpb;visionpb";
 option java_multiple_files = true;
 option java_outer_classname = "GeometryProto";
 option java_package = "com.google.cloud.vision.v1";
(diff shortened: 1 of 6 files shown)


quartzmo:
Moved to vision: missing package paths in googleapis protos for v2 client googleapis/google-cloud-go#7534.

wyattanderson:
> Just curious, what are the particulars of your build that prevent you from using the package distributed with the client?

The issue rears its head when we write our own Protobuf definitions that include the `google.cloud.vision.v1` definitions. Say we have our own message type that refers to a `google.cloud.vision.v1` type: the generated Go code will include imports for `cloud.google.com/go/vision/apiv1/visionpb`, which doesn't exist and can't be resolved.

Our build system uses Bazel, but I think this would occur even without Bazel. We work around the issue by patching the repository, as demonstrated here: https://github.com/wyattanderson/googleapis-reproducer/blob/main/patches/com_google_googleapis-importpath.patch

quartzmo:
@wyattanderson Thanks, it's pretty much what we expected. Great to have that documented here for others who have a similar issue.

quartzmo:
From [wyattanderson/googleapis-reproducer/](https://github.com/wyattanderson/googleapis-reproducer)

> This is an attempt at getting Bazel, Protobuf, and Go to work nicely together without too many hacks or workarounds.

Oops, sorry for making it worse :(

PR #792, opened 2023-04-06
chore: add Go build targets for compute

codyoss:
Note: manually updated proto file, but it yields the same result of the new disco-converter would generate.

Fixes: https://github.com/googleapis/google-cloud-go/issues/7345

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -122,7 +122,7 @@ http_archive(
     urls = ["https://github.com/googleapis/rules_gapic/archive/v%s.tar.gz" % _rules_gapic_version],
 )
 
-# This must be above the download of gRPC (in C++ section) and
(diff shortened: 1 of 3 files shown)


noahdietz (commented):
LGTM but can you revert the WORKSPACE changes? these will just be overwritten by copybara / I don't want to introduce a weird copybara push issue.

codyoss:
@noahdietz done. Although I think the changes there are good?!? At least my editor thinks we should format like that 😝

noahdietz:
> @noahdietz done. Although I think the changes there are good?!? At least my editor thinks we should format like that 😝

They are indeed good, but let's make them in google3 :)

noahdietz (approved):

PR #831, opened 2023-07-09
feat: Adding proto_library for bytestream

linzhp:
Adding proto_library for bytestream so people can import the proto

### DIFF
--- a/google/bytestream/BUILD.bazel
+++ b/google/bytestream/BUILD.bazel
@@ -1,0 +1,11 @@
+load("@rules_proto//proto:defs.bzl", "proto_library")
+
+proto_library(
+    name = "bytestream_proto",
+    srcs = ["bytestream.proto"],
+    visibility = ["//visibility:public"],
+    deps = [
+        "//google/api:annotations_proto",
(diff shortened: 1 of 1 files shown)


sluongng:
@linzhp you might need to generate this file instead of manually create it

```
> bazel run :build_gen -- --src=google/bytestream
```

Not quite sure if the root `api-index-v1.json` file needs to be updated as well?

@parthea do you know how should we go about doing this?

linzhp:
`bazel run :build_gen` would create many targets but we only need the `proto_library` target. I prefer to wait until a real use case to come up to generate those targets.

bitcoinfinancier (approved):

alexander-fenster:
Hi folks,

We cannot accept pull request here because the content of this repository is synced from the internal repository, but I did generate the `BUILD.bazel` for `google/bytestream` in https://github.com/googleapis/googleapis/commit/6ed14f1fb7d111282fd1953441e30f13c4a6d847, hope that works for you!

I'll close this PR, let me know if anything else needs to be fixed.

PR #855, opened 2023-09-27
make googleapis a Bzlmod module

tyler-french:


tyler-french:
@fmeum Could you take a look? It's a bit difficult because this package doesn't have versioned releases

MODULE.bazel
```diff
@@ -0,0 +1,22 @@
+module(
+    name = "com_google_googleapis",
+    version = "0.0.1",
```
fmeum, line 3: You can leave this at `0.0.0` and rely on the Publish to the BCR plugin to patch in the real version number when a tag is created. Well, if no tags are created we will need to release this manually, e.g. using Go pseudo-versions.

MODULE.bazel
```diff
@@ -0,0 +1,22 @@
...
+)
+
+bazel_dep(name = "protobuf", version = "21.7", repo_name = "com_google_protobuf")
+bazel_dep(name = "rules_proto", version = "5.3.0-21.7")
+
+switched_rules = use_extension("//:extensions.bzl", "switched_rules")
+switched_rules.use_languages(
+    cc = True,
```
fmeum, line 11: I would start small here since the way this is implemented every of these language rulesets will need to be in the BCR and a dep of this module to work with strict deps. We could start with `cc`, `java`, `python` and `go` for now.

extensions.bzl
```diff
@@ -0,0 +1,50 @@
...
+)
+
+def _switched_rules_impl(ctx):
+    attrs = {}
+    for module in ctx.modules:
+        if not module.is_root:
+            continue
+
+        for t in module.tags.use_languages:
+            attrs = {
```
fmeum, line 26: This results in latest wins semantics. Should we fail on multiple tags instead?

MODULE.bazel
```diff
@@ -0,0 +1,22 @@
+module(
```
fmeum, line 1: We should enable a CI run with Bzlmod (only the default in the not yet released Bazel 7).

silicon-ninja:
Hey everyone,

I'm interested in helping with the transition of `googleapis` to `bzlmod` support. While reviewing the dependency graph, I noticed a key decision point: should we opt for pure `bzlmod` support, or consider a partial approach using `use_repo`?

Adopting a purely `bzlmod`-based approach means that all `gapic` dependencies would also need to support `bzlmod`. This, in turn, hinges on `grpc` being a `bzlmod` project. However, it appears that `grpc` requires a substantial patch file to support `bzlmod`, which could complicate matters.

What are the team's thoughts on this? Are there any plans already in place for tackling these dependencies?

extensions.bzl
SanjayVas: I'd recommend using a logical OR for each language. That's what I have in my own fork.

alexander-fenster:
Hi @tyler-french, @fmeum, and @silicon-ninja,

Thank you for your suggestions! We indeed cannot immediately switch to bzlmod: gRPC dependency makes it somewhat complicated, but that's not the only issue. For now, we will keep using Bazel 6 (via Bazelisk) and `WORKSPACE` for some more time, while we think internally how we can switch. We definitely should figure out how to switch soon because of Bazel's plans to deprecate / remove `WORKSPACE` support.

Please note that we are unable to merge pull requests here, because all Bazel configuration (and protos) in this repository is auto-synced from the internal code storage. I will leave this PR open for discussions, but the code change will come from the regular sync process, so this PR will not get merged.

Thank you!

arnehormann:
Hi @alexander-fenster, isn't it still possible to support both in a soft migration?
See https://bazel.build/external/migration#workspace.bzlmod - a `WORKSPACE.bzlmod` file that makes Bazel with bzlmod ignore the `WORKSPACE` file while Bazel without bzlmod will ignore `WORKSPACE.bzlmod` and `MODULE.bazel`

SanjayVas:
See https://github.com/bazelbuild/bazel-central-registry/issues/1113

Note that the migration of core ecosystem infrastructure such as this blocks downstream migration. I appreciate in this case that it's in fact upstream dependencies that can block this one, e.g. protobuf and grpc.

My project just uses googleapis for proto definitions in the `google.type` and `google.api` packages, so I was able to patch in just enough support in [my own Bazel registry](https://github.com/SanjayVas/bazel-registry/tree/main/modules/googleapis).

SanjayVas:
Perhaps the [global common component types](https://google.aip.dev/213#existing-global-common-components) should be split out from googleapis.

alexander-fenster:
@SanjayVas We actually have https://github.com/googleapis/api-common-protos for this specific purpose, I'll see if we can actually update _that one_ first and see if it helps.

In general, folks, please give us some time and we'll figure this out - I just want to set some expectations that it likely won't happen until the new year. There are a lot of moving parts involved (much more than I would like to have :) ), but we are on it.

SanjayVas:
> @SanjayVas We actually have https://github.com/googleapis/api-common-protos for this specific purpose, I'll see if we can actually update that one first and see if it helps.

I am aware, but that often lags way behind the main googleapis repo. It's not automatically kept up-to-date. I imagine that's why the AIPs point to googleapis as the canonical location and why lots of core Bazel infra (e.g. rules_go) depends on it.

extensions.bzl
richardburce19:
```suggestion
            attrs = {
```

tyler-french: Opting for a more restrictive approach first, we can adjust later.

MODULE.bazel
tyler-french: I am not sure CI is running in this repo. I will try to set up a BCR presubmit also.

ah-edg:
> @SanjayVas We actually have https://github.com/googleapis/api-common-protos for this specific purpose, I'll see if we can actually update _that one_ first and see if it helps.
> 
> In general, folks, please give us some time and we'll figure this out - I just want to set some expectations that it likely won't happen until the new year. There are a lot of moving parts involved (much more than I would like to have :) ), but we are on it.

Hi @alexander-fenster, I hope it's okay to ping you after 2.5 months.
As it's the new year now, I'd love to get an update if you could please provide one. Even if it's just a rough plan. This issue (here and in grpc / grpc-java) blocks migrations to Bazel 7 and locks us into legacy software and into some maintenance heavy workarounds. I'd really love to get rid of them without additional manual patches.

extensions.bzl
```diff
@@ -0,0 +1,59 @@
...
+    for module in ctx.modules:
+        if not module.is_root:
+            continue
+
+        is_tag_set = False
+        set_tag_name = ""
+
+        for t in module.tags.use_languages:
+            if is_tag_set:
+                fail("Multiple use_language tags are set in the root module: '{}' and '{}'. Only one is allowed.".format(set_tag_name, t.name))
```
keith, line 30:
```suggestion
                fail("Multiple use_language tags are set in the root module: '{}' and '{}'. Only one is allowed.".format(set_tag_name, module.name))
```

extensions.bzl
```diff
@@ -0,0 +1,59 @@
...
+
+            is_tag_set = True
+            set_tag_name = t.name
```
keith, line 33:
```suggestion
            set_tag_name = module.name
```

keith (commented):
`t` doesn't have a `name`, so I assume you meant module here

MODULE.bazel
```diff
@@ -0,0 +1,16 @@
...
+
+bazel_dep(name = "protobuf", version = "21.7", repo_name = "com_google_protobuf")
+bazel_dep(name = "rules_proto", version = "5.3.0-21.7")
+
+switched_rules = use_extension("//:extensions.bzl", "switched_rules")
+switched_rules.use_languages(
+    cc = True,
+    go = True,
+    java = True,
+    python = True,
```
keith, line 14:
```suggestion
    python = True,
    grpc = True,
```

It looks to me if you add this some other deps are missing

keith: maybe blocked on https://github.com/bazelbuild/bazel-central-registry/issues/353

SeeYaaG:
Explain. Ihave no clue but definitely interested

keith:
i stacked some commits on this over here https://github.com/googleapis/googleapis/pull/892

Shikho18 (approved):

Shikho18:
- [ ] - - Shikho18

shivaninischal0:
https://github.com/googleapis/googleapis/pull/956#issuecomment-2785491987

Angelwoolridge (approved):

PR #860, opened 2023-10-13
fix typos

shuoer86:


alexander-fenster:
Thanks @shuoer86, unfortunately, this is not a source of truth but a copy of the internal files that is autosynced, and we cannot accept PRs here. We'll eventually fix those typos internally.

PR #864, opened 2023-11-08
feat: add a new target

JoeWang1127:


alexander-fenster:
Hi @JoeWang1127,

This needs to be applied internally first, then it will be synced here. I'll reach out with more instructions :)

Thanks,
-Alex

PR #873, opened 2024-01-07
Rename service.proto to YourHomeModPro.proto

ArSenic04:


ArSenic04:
feat: Rename file using conventional commit guidelines

PR #878, opened 2024-02-01
fix(diregapic): s/bazel/bazelisk/ in DIREGAPIC build GitHub action 

vchudnov-g:
This seems to be related to the upgrade in `WORKSPACE` to Bazel 6.3.0


parthea (approved):

vchudnov-g:
Adding `DO NOT MERGE` so we remember to take out the extra `pull_request` trigger that we're using to verify this PR. We don't want it in production.

vchudnov-g:
The GitHub action failed with `google/cloud/compute/v1/compute.proto:29536:3: "null" is not defined.` I'll continue debugging. Marking this PR `draft` for now.

PR #892, opened 2024-03-27
feat(bazel): make googleapis a Bzlmod module

keith:
This depends on https://github.com/bazelbuild/bazel-central-registry/pull/1699 because of a circular dependency on grpc-java

This is based on https://github.com/googleapis/googleapis/pull/855 with some fixes

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -1,16 +1,29 @@
 module(
-    name = "com_google_googleapis",
-    version = "0.0.0",
+    name = "googleapis",
+    version = "0",
+    repo_name = "com_google_googleapis",
 )
 
+bazel_dep(name = "grpc", version = "1.56.3.bcr.1", repo_name = "com_github_grpc_grpc")
+bazel_dep(name = "grpc-java", repo_name = "io_grpc_grpc_java")
 bazel_dep(name = "protobuf", version = "21.7", repo_name = "com_google_protobuf")
+bazel_dep(name = "rules_go", version = "0.46.0", repo_name = "io_bazel_rules_go")
 bazel_dep(name = "rules_proto", version = "5.3.0-21.7")
+bazel_dep(name = "rules_python", version = "0.31.0")
 
 switched_rules = use_extension("//:extensions.bzl", "switched_rules")
+
+# TODO: Enable support for other languages with bzlmod
 switched_rules.use_languages(
     cc = True,
+    # csharp = True,
+    # gapic = True,
     go = True,
+    grpc = True,
     java = True,
+    # nodejs = True,
+    # php = True,
     python = True,
+    # ruby = True,
 )
 use_repo(switched_rules, "com_google_googleapis_imports")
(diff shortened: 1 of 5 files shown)


MODULE.bazel
```diff
@@ -0,0 +1,29 @@
...
+bazel_dep(name = "rules_proto", version = "5.3.0-21.7")
+bazel_dep(name = "rules_python", version = "0.31.0")
+
+switched_rules = use_extension("//:extensions.bzl", "switched_rules")
+
+# TODO: Enable support for other languages with bzlmod
+switched_rules.use_languages(
+    cc = True,
+    # csharp = True,
+    # gapic = True,
```
keith, line 20: i left some of these out, which means not everything in the repo builds with bzlmod enabled, in order to support each of these we'll have to move their deps to bzlmod, which is potentially difficult depending on their transitives

WORKSPACE.bzlmod
```diff
@@ -0,0 +1,23 @@
+# When Bzlmod is enabled, this file replaces the content of the original WORKSPACE
+# and makes sure no WORKSPACE prefix or suffix are added when Bzlmod is enabled.
+
+load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
+
+_disco_to_proto3_converter_version = "74c0ede1e6871c878de6c3cab3ef152280c6518f"
```
meteorcloudy, line 6: Are those only dev dependencies?

meteorcloudy (commented):
Nice, this a great start!

WORKSPACE.bzlmod
keith: they aren't, but i didn't handle all the different language specific dependencies here, and this one is one that was easy enough to throw here so that a few more targets built when testing the integration in this repo.

meteorcloudy (approved):

keith:
@meteorcloudy could you help route this?

meteorcloudy:
I'll ask around

michaelschuett-tomtom:
Thank you for this. Saved me a ton of time instead of trying to figure out how to port switched_rules over. I feel like that should be documented some place although it's not documented for use in WORKSPACE as well so not really regressing at all here.

keith:
yea i think if this gets merged we can improve these things. this is already included in the BCR and there are a few examples in there like grpc-java using this extension

meteorcloudy:
@wyverald will follow this up with the internal teams. We should really improve the protobuf + grpc + googleapis situation.

mering:
@meteorcloudy any updates?

Wyverald:
Hello -- I'm still in talks with the team internally at Google. There's a concurrent proposal to reshape the project structure or something, and they're apparently exploring other options of exposing the protos to downstream users.

neilconway:
@Wyverald Any news on this?

loeffel-io:
+1

meteorcloudy:
See more details in https://github.com/bazelbuild/bazel-central-registry/pull/3472, we believe the new approach will simplify the use of googleapis with Bzlmod and avoid polluting your dependency graph.

PR #988, opened 2025-02-28
build_events.proto: add support for Buck2 events

sluongng:
As discussed via email, we would like to extend the existing Build Event
Protocol and Service definitions for Buck2 event system as well.


michaeledgar:
Re-posting from an email conversation:

> Having worked on the effort that first brought these APIs into public use, my perspective is: it would be great to have other build tools using this general-purpose transfer protocol! That was actually an unmet goal of that project 6 years ago. Would it be possible to reuse the existing `Any` fields for the Buck stream, and indicate in metadata that the stream is a Buck stream (vs. Bazel)?

My reasoning has a few steps.

1. The `bazel_event` field was always a misnomer -- even when first introduced, the field was also used for `blaze`, the Google-internal variant of Bazel. So as annoying as it is, that field should be presumed to have the semantics of a hypothetical field generically named `build_tool_event`.
2. The existing fields are already `google.protobuf.Any` -- so it is permissible for a stream to use the `BuildEvent.bazel_event` field with arbitrary message types. While Bazel and Google's `blaze` use `build_event_stream.BuildEvent` (Bazel's event type), it is valid to instead supply `buck2.BuildEvent` in this field. Clients already need to downcast the field to a given event type.
3. In fact, a single stream could contain different types for different events -- if you have a hypothetical franken-build that uses both `bazel` and `buck2`, it is permissible to smash their combined build events into a single stream, with each tool populating the `BuildEvent.bazel_event` field. It might be challenging to assign sequence numbers correctly, but that's orthogonal to the question of which field should be used.
4. We can't justify adding a new field for every possible build tool -- which would make the grade?

One could reasonably read these arguments and respond: "But Mike, what about `google.protobuf.Any BazelEvent.build_execution_event`? What about `google.protobuf.Any BazelEvent.source_fetch_event`? There are other `Any` fields for other purposes. It seems to make sense we would add a new `Any` field for a different build tool."

That is an excellent argument -- and all I can say is: I was the TL of the team that introduced those fields, and they were a mistake. We should have done a bit more legwork to squeeze new events into the `bazel_event` field. We were working on an experimental product and chose the path of least resistance. Putting our new events in the `bazel_event` field would have meant changing some of the event consumers in Google's infrastructure to verify the event types before downcasting `bazel_event` to `build_event_stream.BuildEvent`, and we were lazy and didn't want to do that, thinking this would be "wasted effort" if the experiment didn't launch. It would have been a handful of tiny changes but we took a shortcut instead. It really was just poor judgment. In hindsight - thankfully, nothing launched using those fields, and if it's possible we should mark them deprecated and discourage their use. The mistake we made was thinking it would have been "wasted effort" to make google-internal infrastructure tolerate other event types. Tolerating arbitrary event types without crashing/failing is actually required for any correct event consumer. (We also overestimated how quickly we needed to change the event producers and underestimated how long it would take to update the consuming services.)

I recognize that the rationale I'm now calling a "shortcut" and "lazy" is perhaps relevant to the current discussion, and I want to be clear that my disparagement is reserved for myself, somebody who was very close to the origins of this API and could have consulted with anybody affected at a moment's notice.

The right thing to do for *any* consumer of the `BuildEvent.bazel_event` field is to check the type before downcasting, to tolerate messages of unknown types without failure, and to handle only the event types the consumer understands. That allows any producer to begin populating the field with new types, and for new consumers to be written that understand the new types, all without changing the actual API schema itself. That's what I think buck2 should do, and any other tool.

sluongng:
@michaeledgar that sounds good to me. I think it would be nicer if we can use a generic `build_tool_event` instead of the existing `bazel_event` field to better communicate the intention, but that's mostly cosmetic.

I have adjusted this PR accordingly.

cc: @TheGrizzlyDev from EngFlow and @iguridi @alexlian from Meta

bduffany:
> I think it would be nicer if we can use a generic build_tool_event instead of the existing bazel_event field to better communicate the intention, but that's mostly cosmetic.

I think it would be really unfortunate to have such a core part of the spec have such an unclear name. It would be like having the "HTTP request body" be called "HTTP request UFO"

michaeledgar:
I am sending out for review an internal change to this API that will add the new `BuildTool.build_tool_event` field. It will be a while before Bazel uses this by default, but once the new field is available in BES API servers, I can introduce a flag in Bazel to switch to using the new field.

sluongng:
@michaeledgar friendly ping on `build_tool_event` field. 

It seems like https://github.com/buildbarn/bonanza will be another potential candidate for using this event system.

2FaceS-bit (approved):

2FaceS-bit (approved):

sluongng:
@michaeledgar any chance of this happening still? If this goes inactive for too long, we would have to explore alternative routes with Buck2 (and other build tools) telemetry events.

sluongng:
@michaeledgar I have created https://github.com/facebook/buck2/pull/995 as a result of this conversation. I hope the actual change can be made upstream inside google3 so the OSS ecosystem can benefit from existing server implementations.

2FaceS-bit (approved):

PR #993, opened 2025-03-13
build: deploy use of converter config in GCE BUILD rule + DIREGAPIC action

vchudnov-g:
This will start working once the `WORKSPACE` file is updated to or past [this one](https://github.com/googleapis/disco-to-proto3-converter/commit/ebc9e9e267bfbc9a1173dd5f212a2dd03bf7b9d5); the source-of-truth of that is Google-internal and gets propagated out.

### DIFF
--- a/google/cloud/compute/v1/compute.config.json
+++ b/google/cloud/compute/v1/compute.config.json
@@ -1,0 +1,13 @@
+{
+  "converterVersion": "UNSPECIFIED",
+  "apiVersion": "",
+  "discoveryRevision": "",
+  "inlineSchemas" : [{
+    "schema": "",
+    "locations": {
+      "ErrorSet": [
+        "schemas.ManagedInstanceLastAttempt.errors"
+      ]
+    }
+  }]
+}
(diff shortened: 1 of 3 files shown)


google/cloud/compute/v1/compute.config.json
```diff
@@ -0,0 +1,13 @@
+{
+  "converterVersion": "UNSPECIFIED",
+  "apiVersion": "",
+  "discoveryRevision": "",
+  "inlineSchemas" : [{
+    "schema": "",
+    "locations": {
+      "ErrorSet": [
```
vchudnov-g, line 8: Please confirm this is a reasonable set for this inline schema that was causing the original problem.

chingor13 (approved):
Stamping this to unblock. I can't answer your question, so perhaps get confirmation before merging

vchudnov-g:
Adding `do not merge` until we get confirmation on the config from the GCE team.

vchudnov-g:
Thanks!

PR #1005, opened 2025-03-29
feat(http): add support for mapping headers to request fields

toumorokoshi:
Hello! This is primarily a way to start a conversation, but I wanted to illustrate a somewhat clear proposal as well.

This proposes a method to map HTTP headers to requests and  responses in a protobuf request.

This is useful for implementing IETF RFCs (such as 9110's if-match header) in HTTP+JSON apis that 
are exposed in gRPC-JSON transcoding.

This was also brought up in a GitHub discussion: (see https://github.com/googleapis/googleapis/discussions/749)

### DIFF
--- a/google/api/http.proto
+++ b/google/api/http.proto
@@ -45,17 +45,18 @@ message Http {
 // gRPC Transcoding is a feature for mapping between a gRPC method and one or
 // more HTTP REST endpoints. It allows developers to build a single API service
 // that supports both gRPC APIs and REST APIs. Many systems, including [Google
-// APIs](https://github.com/googleapis/googleapis),
-// [Cloud Endpoints](https://cloud.google.com/endpoints), [gRPC
-// Gateway](https://github.com/grpc-ecosystem/grpc-gateway),
-// and [Envoy](https://github.com/envoyproxy/envoy) proxy support this feature
-// and use it for large scale production services.
+// APIs](https://github.com/googleapis/googleapis), [Cloud
+// Endpoints](https://cloud.google.com/endpoints), [gRPC
+// Gateway](https://github.com/grpc-ecosystem/grpc-gateway), and
+// [Envoy](https://github.com/envoyproxy/envoy) proxy support this feature and
+// use it for large scale production services.
 //
 // `HttpRule` defines the schema of the gRPC/REST mapping. The mapping specifies
 // how different portions of the gRPC request message are mapped to the URL
-// path, URL query parameters, and HTTP request body. It also controls how the
-// gRPC response message is mapped to the HTTP response body. `HttpRule` is
-// typically specified as an `google.api.http` annotation on the gRPC method.
+// path, URL query parameters, HTTP headers, and HTTP request body. It also
+// controls how the gRPC response message is mapped to the HTTP response body.
+// `HttpRule` is typically specified as an `google.api.http` annotation on the
+// gRPC method.
 //
 // Each mapping specifies a URL path template and an HTTP method. The path
 // template may refer to one or more fields in the gRPC request message, as long
@@ -85,8 +86,8 @@ message Http {
 // - gRPC: `GetMessage(name: "messages/123456")`
 //
 // Any fields in the request message which are not bound by the path template
-// automatically become HTTP query parameters if there is no HTTP request body.
-// For example:
+// nor specified as values in `request_headers` automatically become HTTP query
+// parameters if there is no HTTP request body. For example:
 //
 //     service Messaging {
 //       rpc GetMessage(GetMessageRequest) returns (Message) {
@@ -112,14 +113,13 @@ message Http {
 //
 // Note that fields which are mapped to URL query parameters must have a
 // primitive type or a repeated primitive type or a non-repeated message type.
-// In the case of a repeated type, the parameter can be repeated in the URL
-// as `...?param=A&param=B`. In the case of a message type, each field of the
+// In the case of a repeated type, the parameter can be repeated in the URL as
+// `...?param=A&param=B`. In the case of a message type, each field of the
 // message is mapped to a separate parameter, such as
 // `...?foo.a=A&foo.b=B&foo.c=C`.
 //
-// For HTTP methods that allow a request body, the `body` field
-// specifies the mapping. Consider a REST update method on the
-// message resource collection:
+// For HTTP methods that allow a request body, the `body` field specifies the
+// mapping. Consider a REST update method on the message resource collection:
 //
 //     service Messaging {
 //       rpc UpdateMessage(UpdateMessageRequest) returns (Message) {
@@ -134,17 +134,15 @@ message Http {
 //       Message message = 2;   // mapped to the body
 //     }
 //
-// The following HTTP JSON to RPC mapping is enabled, where the
-// representation of the JSON in the request body is determined by
-// protos JSON encoding:
+// The following HTTP JSON to RPC mapping is enabled, where the representation
+// of the JSON in the request body is determined by protos JSON encoding:
 //
 // - HTTP: `PATCH /v1/messages/123456 { "text": "Hi!" }`
 // - gRPC: `UpdateMessage(message_id: "123456" message { text: "Hi!" })`
 //
-// The special name `*` can be used in the body mapping to define that
-// every field not bound by the path template should be mapped to the
-// request body.  This enables the following alternative definition of
-// the update method:
+// The special name `*` can be used in the body mapping to define that every
+// field not bound by the path template should be mapped to the request body.
+// This enables the following alternative definition of the update method:
 //
 //     service Messaging {
 //       rpc UpdateMessage(Message) returns (Message) {
@@ -165,14 +163,14 @@ message Http {
 // - HTTP: `PATCH /v1/messages/123456 { "text": "Hi!" }`
 // - gRPC: `UpdateMessage(message_id: "123456" text: "Hi!")`
 //
-// Note that when using `*` in the body mapping, it is not possible to
-// have HTTP parameters, as all fields not bound by the path end in
-// the body. This makes this option more rarely used in practice when
-// defining REST APIs. The common usage of `*` is in custom methods
-// which don't use the URL at all for transferring data.
+// Note that when using `*` in the body mapping, it is not possible to have HTTP
+// parameters, as all fields not bound by the path end in the body. This makes
+// this option more rarely used in practice when defining REST APIs. The common
+// usage of `*` is in custom methods which don't use the URL at all for
+// transferring data.
 //
-// It is possible to define multiple HTTP methods for one RPC by using
-// the `additional_bindings` option. Example:
+// It is possible to define multiple HTTP methods for one RPC by using the
+// `additional_bindings` option. Example:
 //
 //     service Messaging {
 //       rpc GetMessage(GetMessageRequest) returns (Message) {
@@ -203,18 +201,15 @@ message Http {
 //    message) are classified into three categories:
 //    - Fields referred by the path template. They are passed via the URL path.
 //    - Fields referred by the [HttpRule.body][google.api.HttpRule.body]. They
-//    are passed via the HTTP
-//      request body.
+//    are passed via the HTTP request body.
 //    - All other fields are passed via the URL query parameters, and the
 //      parameter name is the field path in the request message. A repeated
 //      field can be represented as multiple query parameters under the same
 //      name.
 //  2. If [HttpRule.body][google.api.HttpRule.body] is "*", there is no URL
-//  query parameter, all fields
-//     are passed via URL path and HTTP request body.
+//  query parameter, all fields are passed via URL path and HTTP request body.
 //  3. If [HttpRule.body][google.api.HttpRule.body] is omitted, there is no HTTP
-//  request body, all
-//     fields are passed via URL path and URL query parameters.
+//  request body, all fields are passed via URL path and URL query parameters.
 //
 // Path template syntax
 //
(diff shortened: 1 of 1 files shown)


google/api/http.proto
```diff
@@ -45,17 +45,18 @@ message Http {
 // gRPC Transcoding is a feature for mapping between a gRPC method and one or
 // more HTTP REST endpoints. It allows developers to build a single API service
 // that supports both gRPC APIs and REST APIs. Many systems, including [Google
-// APIs](https://github.com/googleapis/googleapis),
-// [Cloud Endpoints](https://cloud.google.com/endpoints), [gRPC
```
rofrankel, line 49: Nit:  this diff would be much easier to evaluate without spurious formatting changes.

google/api/http.proto
```diff
@@ -346,6 +339,18 @@ message HttpRule {
   // message type.
   string body = 7;
 
+  // The name of fields whose values are mapped to HTTP headers in a request. If
+  // a header has a case-insensitive match to a key in this map, the field in
+  // the request that matches the value will have it's value set with the value
```
rofrankel, line 344:
```suggestion
  // the request that matches the value will have its value set to the value
```

google/api/http.proto
```diff
@@ -346,6 +339,18 @@ message HttpRule {
   // message type.
   string body = 7;
 
+  // The name of fields whose values are mapped to HTTP headers in a request. If
```
rofrankel, line 342:
```suggestion
  // A mapping from HTTP headers to request fields. If
```

The text this would replace is describing the keys of the map, rather than the map itself.  It's also a little redundant with the second sentence.

google/api/http.proto
```diff
@@ -346,6 +339,18 @@ message HttpRule {
...
+  // a header has a case-insensitive match to a key in this map, the field in
```
rofrankel, line 343: What happens if multiple map values refer to the same field?  Would it make more sense to invert the mapping so that request fields rather than headers benefit from the uniqueness of keys?

I can't think of a realistic use case for mapping the same header to multiple fields, but it still seems better semantically than mapping multiple headers to the same field (which might even have undefined behavior if not otherwise specified here, e.g. given a naive implementation that effectively depends on the (undefined) ordering of map entries).

google/api/http.proto
```diff
@@ -346,6 +339,18 @@ message HttpRule {
...
+  // the request that matches the value will have it's value set with the value
+  // in the header.
+  //
+  // If a field is specified in this map, it will not be mapped to a query
+  // parameter.
+  //
+  // NOTE: the referred field must be present at the top-level of the request
```
rofrankel, line 350:
```suggestion
  // NOTE: the field must be a top-level field in the request
```

Phrasing nits, feel free to reject or adapt as you see fit.

rofrankel, line 350: Why is it a requirement that the field be top-level?  Is there a reason to prohibit `.`-delimited field paths?

google/api/http.proto
```diff
@@ -354,6 +359,15 @@ message HttpRule {
   // message type.
   string response_body = 12;
 
+  // The name of fields whose values are mapped to HTTP headers in a response.
```
rofrankel, line 362:
```suggestion
  // A mapping from response fields to HTTP headers in the response.
```

google/api/http.proto
```diff
@@ -354,6 +359,15 @@ message HttpRule {
...
+  // If a field is in the response that matches a value in this map, an HTTP
+  // response header with the name of the key in the map will be populated
+  // with that value.
+  //
+  // NOTE: the referred field must be present at the top-level of the request
```
rofrankel, line 367:
```suggestion
  // NOTE: the referred field must be present at the top-level of the request
```

(same two comments from above apply here)

google/api/http.proto
```diff
@@ -354,6 +359,15 @@ message HttpRule {
   // message type.
   string response_body = 12;
 
+  // The name of fields whose values are mapped to HTTP headers in a response.
+  // If a field is in the response that matches a value in this map, an HTTP
```
rofrankel, line 363: Same comment as above, but inverted:  what if multiple fields are mapped to the same header?  (The use case for mapping the same field value to multiple headers also seems like a stretch, but at least the behavior is better defined...)

rofrankel (changes requested):
+100 for the spec having a way to express this, thanks for drafting a proposal @toumorokoshi!

Added some comments on the details.

google/api/http.proto
wora: We should limit the complexity of initial implementation. If there is real need, we can relax the rule later.

google/api/http.proto
wora: This issue already exists today. For example, ?a=1&b=2&a=3&b=4

The current implementation, hence the semantics, is the last one wins.

rofrankel: Query params are ordered, so "last one wins" is well-defined.  But map entries are not ordered, so even if we wanted "last one wins" semantics, I'm not sure we could have them.

Inverting the maps seems to make this problem go away entirely.  Do you see any downsides to doing so?

google/api/http.proto
rofrankel: That's definitely a valid approach.  But iterating on the spec like this could be costly in terms of adoption and fragmentation.

For example, if we define the spec in this more limited way, and then Microsoft adds support for it, and then we expand the spec, Microsoft may take some amount of time to make another update (and even if they're quick about it, the changes could end up on either sides of a .NET version boundary).

And in parallel, some other implementation might only add support for header mapping once we eventually add full field paths to the spec.

So we could end up in a world where different implementations (or versions thereof) support header mappings in different ways, making the ecosystem a little more confusing for everyone.

I suspect that implementing this kind of field path traversal is probably not particularly expensive in the context of a full transcoding implementation.  So my gut says that the savings of a simpler initial spec change are probably not worth the potential overhead.  (And if it does turn out to be a problematic cost for a given implementation, that implementation could always (temporarily) support only the subset of the spec that doesn't allow field path traversal for header mappings, and could document its limitation.)

WDYT?

google/api/http.proto
```diff
@@ -346,6 +339,18 @@ message HttpRule {
   // message type.
   string body = 7;
 
+  // The name of fields whose values are mapped to HTTP headers in a request. If
+  // a header has a case-insensitive match to a key in this map, the field in
+  // the request that matches the value will have it's value set with the value
+  // in the header.
+  //
+  // If a field is specified in this map, it will not be mapped to a query
+  // parameter.
```
wora, line 348: The current implementation already allows a parameter be duplicated by body and HTTP query parameters. We cannot suddenly block a parameter from query parameter. That is a break changing.

google/api/http.proto
wora: That makes sense. In that case, we should support whatever is supposed by {param} syntax.

google/api/http.proto
wora: Can you spell out the details and let's see if it can fully address ambiguity?

google/api/http.proto
toumorokoshi: I don't think either is realistic - both keys and values should just be unique, so I think i'll just update it to state as such?

I can't think of a good reason to allow either multiple headers mapping to the same field, or multiple fields mapping to the same header.

google/api/http.proto
toumorokoshi: I think it's better to just state that the keys and values must all be unique here, allowing only 1-1 mappings. 

Is there really a good reason for allowing many-to-one in either direction? I can't think of one that wouldn't just lead to more esoteric behavior (e.g. if there's two mappings, which one wins?)

google/api/http.proto
toumorokoshi: updated to make them all unique, and we can carry on the larger discussion in the other thread.

google/api/http.proto
toumorokoshi: Updated to require all keys/values to be unique. thoughts?

google/api/http.proto
toumorokoshi: If an existing field was added to this map, then yes, I think that would be a breaking change. I imagined that, even once this mapping exists, then existing fields would not map to it.

I don't think the alternative is desirable: that would imply that any field not in the body would always be able to be provided as a query parameter, and then optionally a header parameter (by adding it to the map).

There would be no way to specify that a field is exclusively honored in the header.

google/api/http.proto
toumorokoshi: updated!

toumorokoshi:
@jskeet @noahdietz hi! we were discussing offline and felt you might be the right people to loop into this discussion.

We feel that a way to enable proto-HTTP header mappings is necessary for enabling gRPC APIs which can be transcoded to those that adhere to IETF RFCs (such as [9110's if-match](https://www.rfc-editor.org/rfc/rfc9110.html#name-preconditions)). 

This PR itself probably will not got anywhere, but I'm wondering if you have suggestions on how we could move this conversation forward.

noahdietz:
Hi hi!

> @jskeet @noahdietz hi! we were discussing offline and felt you might be the right people to loop into this discussion.

I've engaged with the folks that own this proto and the transcoding infra and will hopefully get someone from there involved.

> This PR itself probably will not got anywhere

Right, as you know, we'd need to update the internal proto and mirror it here.

> I'm wondering if you have suggestions on how we could move this conversation forward.

Let's keep discussing here, and I will try to get others to weigh in.

google/api/http.proto
rofrankel: There's protocol-level enforcement of key uniqueness, but value uniqueness needs to be implemented separately (e.g. by a linter).  Not the end of the world, of course.

> Is there really a good reason for allowing many-to-one in either direction? I can't think of one that wouldn't just lead to more esoteric behavior (e.g. if there's two mappings, which one wins?)

If you allow a header to be mapped to multiple request fields, or a response field to be mapped to multiple headers, then doesn't the "if there's two mappings, which one wins?" question go away?

toumorokoshi:
> Hi hi!
> 
> > @jskeet @noahdietz hi! we were discussing offline and felt you might be the right people to loop into this discussion.
> 
> I've engaged with the folks that own this proto and the transcoding infra and will hopefully get someone from there involved.
> 
> > This PR itself probably will not got anywhere
> 
> Right, as you know, we'd need to update the internal proto and mirror it here.
> 
> > I'm wondering if you have suggestions on how we could move this conversation forward.
> 
> Let's keep discussing here, and I will try to get others to weigh in.

Awesome! thank you for the help.

noahdietz:
> This was also brought up in a GitHub discussion: (see https://github.com/googleapis/googleapis/discussions/749)

@toumorokoshi I'm trying to get a better understanding of where this will ultimately be used. Have you contacted Envoy or grpc-gateway, two of the major OSS projects that implement grpc-HTTP transcoding, yet?

I'm not getting much traction myself and of course most teams are already committed to work for the foreseeable future...

toumorokoshi:
@noahdietz got it! Thanks for pushing.

Our main use case is to adopt IETF RFCs, specifically things like If-Match. https://datatracker.ietf.org/doc/html/rfc7232#section-3.1. We plan to recommend it for aep.dev: https://github.com/aep-dev/aeps/issues/289

We'll reach out to envoy and grpc-gateway too - thanks again!

toumorokoshi:
The level of effort required to support this change in the aformentioned codebases comes down to modifying these two repositories:

- https://github.com/grpc-ecosystem/grpc-httpjson-transcoding for envoy.
- https://github.com/grpc-ecosystem/grpc-gateway for grpc-gateway (it looks like modifying the code generation).

So it's not an intractable amount to get it working. Regarding not getting traction: @noahdietz is it possible to discuss this live with some group? I'd be great to see if there's some middle ground we can find here: both projects have said that having the annotation supported in google.api.http is a baseline requirement.

google/api/http.proto
```diff
@@ -346,6 +347,20 @@ message HttpRule {
   // message type.
   string body = 7;
 
+  // A mapping from HTTP headers to request fields. If
```
wora, line 350: A mapping from HTTP request headers to gRPC request fields. The map key corresponds to an HTTP header name (case-insensitive), and the map value corresponds to an gRPC request field path (using the same syntax as the path template).

google/api/http.proto
```diff
@@ -346,6 +347,20 @@ message HttpRule {
...
+  // a header has a case-insensitive match to a key in this map, the field in
+  // the request that matches the value will have its value set to the value
```
wora, line 352: that matches the value => the gRPC request field that is specified by the map value will ...

google/api/http.proto
```diff
@@ -346,6 +347,20 @@ message HttpRule {
...
+  // in the header.
```
wora, line 353: to the value in the header => to the value of the HTTP header.

We need to specify the precise encoding for string, bytes, number, etc.

google/api/http.proto
```diff
@@ -346,6 +347,20 @@ message HttpRule {
...
+  //
+  // All keys and values of this map **must** be unique.
+  //
+  // If a field is specified in this map, it will not be mapped to a query
```
wora, line 357: a field => a request field

google/api/http.proto
wora: will not be => must not be

Note: This doesn't work for Google APIs. Every Google API request field can be specified by an HTTP request parameter, regardless whether it is mentioned by google.api.http. Please verify other implementations of gRPC transcoding. The current implementation is copy every value from request parameters into body blindly.

google/api/http.proto
```diff
@@ -346,6 +347,20 @@ message HttpRule {
...
+  // parameter.
+  //
+  // The referred field is specified using the field path syntax in a path
```
wora, line 360: I don't think this is necessary if we adopt the suggestion above.

google/api/http.proto
```diff
@@ -354,6 +369,17 @@ message HttpRule {
   // message type.
   string response_body = 12;
 
+  // A mapping from response fields to HTTP headers in the response. If a field
```
wora, line 372: Similar comments as above.

maxgtr816 (approved):

maxgtr816 (approved):

PR #1063, opened 2025-07-03
feat(bazel): automate publications to BCR

diegomarquezp:
_This is a POC and WIP. Do not review or merge._

This PR adds a `.bzl` folder as per the [bazel-contrib instructions](https://github.com/bazel-contrib/publish-to-bcr/tree/main/templates).

The setup was proven correct in https://github.com/bazelbuild/bazel-central-registry/pull/5074

This was done only by using tags (no releases). This is not recommended as releases are recommended (see [presubmit job output](https://buildkite.com/bazel/bcr-presubmit/builds/15951/steps/canvas?sid=0197d201-98a0-4b66-b1dc-936eaf5f4a96)).

```
BcrValidationResult.FAILED: googleapis@0.0.0-20250703-hash-7 is using an unstable source url: `https://github.com/diegomarquezp/googleapis/archive/0.0.0-20250703-hash-7.zip`.
--
  | If at all possible, you should use a release archive URL in the format of `https://github.com/<ORGANIZATION>/<REPO>/releases/download/<version>/<name>.tar.gz` to ensure the archive checksum stability.
  | See https://blog.bazel.build/2023/02/15/github-archive-checksum.html for more context.
  | If no release archives are available, please add a comment to your BCR PR with the text
  | @bazel-io skip_check unstable_url
  | and this check will be skipped.


```


.bcr/patches/module_dot_bazel.patch
mering, line 1: Nit: You can use an overlay instead of a patch for this, too.

MODULE.bazel
```diff
@@ -0,0 +1,21 @@
+module(
+    name = "googleapis",
+    version = "0.0.0-20250703-hash-7",
```
mering, line 3: On main, one usually removes the version attribute and only patched it in the publish action.

PR #1100, opened 2025-09-11
Update account_budget_proposal_service.proto

glav-git:
Just a small typo


jskeet:
The files in this repository are published automatically. In this case, the "iff" may not actually be a typo - it's a common abbreviation for "if and only if". (I don't know whether that's intended in this case.)

If you still feel this is an issue, please raise it with the Ads team. Filing an issue in https://github.com/googleads/google-ads-java would be reasonable, for example.

